PR #1, opened 2016-06-25
Add browser based build, and normalize the `replace` transition.

taras:


taras:
PR for #2

src/list.js
```diff
@@ -7,49 +7,65 @@ export default Primitive.extend('List', {
...
-  fill() {
-    return this.replace(this.value.slice().fill(...arguments));
-  },
-
-  pop() {
-    return this.slice(0, -1);
+  prototype: {
+    forEach: { value() { return this.value.forEach(...arguments); }},
+    valueOf: { value() { return this.value.valueOf(); }}
   },
```
cowboyd: I'm guessing we can get rid of the whole `prototype` hash since we're now storing the transitions inside its own hash so it will never conflict.

src/microstate.js
```diff
@@ -24,30 +26,56 @@ export default class MicroState {
...
+          let transitions = assign({}, descriptor.value, {
+            replace(current, ...args) {
+              return new this.constructor(...args);
+            }
+          });
+          getOwnPropertyNames(transitions).forEach(function(name){
+            let descriptor = getOwnPropertyDescriptor(transitions, name);
+            defineProperty(prototype, name, makeTransitionDescriptor(descriptor));
+          });
         } else if (name === 'valueOf' || name === 'toString') {
```
cowboyd: Once we remove the the `prototype` property, now that we have the `transitions` hash, I think we can get rid of this arm of the `if`. I think this is the only thing keeping this from getting merged.

src/microstate.js
```diff
@@ -24,30 +26,56 @@ export default class MicroState {
...
+          // property is a special method
           defineProperty(prototype, name, descriptor);
+        } else if (descriptor.value && MicroState.isPrototypeOf(descriptor.value)) {
+          // property is a substate
+          
+          defineProperty(prototype, name, MicroState.substate(this, descriptor.value));
```
cowboyd: I assume this is going to be a todo?

Issue #4, opened 2016-07-04
Composable Microstates


cowboyd:
Microstates lets you define atoms of data and the state transitions that move you from one valid atom to the next. Hence we have `number`s, `list`s, `string`s, etc.. However, for more complex interactions, we need to be able to compose these atomic microstates into larger microstates, for example a list of numbers, or a object containing lists of more objects.

We need a way to compose microstates into higher-order objects while still making the API to work with them simple and intuitive. If we lose the intuitiveness of it, then what's the point?

### Simple Construction

It must be simple to construct a composed microstate. That means being able to use a pojo to construct it, and to have it "just work". E.g.

```javascript
let lobs = new ListOfBooleans([true, false, true, true]);
```
You shouldn't have to have any knowledge about the internal format, but be secure that the individual booleans will be passed to the `BooleanState` constructor. Among other things, this allows for simple re-hydration of a microstate from persistent data, or something brought in over the wire.

How would we declare this? Probably something like:

```javascript
import { list, boolean } from 'microstates';

export default list(boolean) {
  transitions: {
    selectAll(current) {
      return current.map(bool => false)
    },
    deselectAll(current) {
      return current.map(bool => true)
    }
  }
}
```
In this case the `list` function takes a constructor function and returns a list constructor that passes each of its constitutents to the boolean constructor.

It is possible that we could implement this by having microstate constructors return instances when invoked with `new` but constructors when invoked without.

In the `selectAll` and `deselectAll` transitions, it should also be implicit that the bare `false` and `true` values will get passed to the `BooleanState` constructor.

### Recursive transitions

In this example, the booleans should not behave like mere booleans. Instead, we should be able to enact all of the normal boolean transitions, but instead of returning the next boolean state, it returns the next state of the entire datastructure that contains them. E.g.

```javascript
let lobs = new ListOfBooleans([true, false, false]);
let next = lobs.get(2).replace(true);
```

which will return a _new_ `ListOfBooleans` object that contains that boolean value replaced with true instead of false.

### Taking Snapshots

Finally, to make things as simple as possible, it should be trivial to get an entire tree of a microstate as a simple object. This will make persistence a snap. The rule to adhere to is that the output of this snapshot should always be valid as the input to the constructor.

my current thinking is that this should be implemented with `valueOf()` which will recurse through all substates and call valueOf() on them. E.g.

```javascript
let lobs = new ListOfBooleans([false, true]);
lobs.valueOf() //=> [false, true]
```

[...]

PR #5, opened 2016-07-08
WIP: Composability Take 2

taras:
This PR is the spike towards #4. I got far enough to see some of the obstacles that we might want to discuss before moving forward.

1. It takes substantial cognitive effort to keep track of what's happening when a *class is defined* vs *in runtime* when transitioning state. Any effort that we put towards making it clear what's executed in design time vs run time will go a long way towards making the code easier to understand and evolve. 

2. Having constructors that transform user input to properties that are merged into the object complicated implementation considerably. It also makes it difficult to reason about how data should be formatted to be merged into state. I found it a lot easier to have no constructors for nested state. We might want to eliminate constructors to keep things simple.

3. We need a clear way to know what needs to be merged. I tried deducing it from target object, but it's difficult to know because some properties are on the prototype and others on the object. Prototype chain complicates this further because we need to accumulate all of the properties from the prototype chain. 

4. I'm using hidden property on the object to start instance of micro states, but these properties are not really hidden. Ideally, we'd keep these in a closure but it's difficult to do programmatically. 

5. One possible solution could be to have a babel plugin that takes a Microstate class and generates as ES6 class. It would give us an extra build step that could make 1-4 easier.

### DIFF
--- a/src/utils/prototype-for.js
+++ b/src/utils/prototype-for.js
@@ -1,0 +1,74 @@
+import assign from './assign';
+import MicroState from '../microstate';
+
+const { getOwnPropertyNames, getOwnPropertyDescriptor, defineProperty } = Object;
+
+export default function prototypeFor(prototype, attrs) {
+
+    return getOwnPropertyNames(attrs).reduce(function(prototype, name) {
+
+      if (name === 'constructor' || name === 'prototype') {
+        return prototype;
+      }
+
+      let descriptor = getOwnPropertyDescriptor(attrs, name);
+
+      if (descriptor.get && typeof descriptor.get === 'function') {
+
+        defineProperty(prototype, name, descriptor);
+
+      } else if (name === 'transitions' && typeof descriptor.value === 'object') {        
+
+        defineTransitions(prototype, descriptor.value);
+
+      } else if (name === 'valueOf' || name === 'toString') {
+        // property is a special method
+        defineProperty(prototype, name, descriptor);
+
+      } else if (descriptor.value && MicroState.isPrototypeOf(descriptor.value)) {
+        // property is a substate      
+        // TODO: apply substate
+
+      } else if (descriptor.value && typeof descriptor.value === 'function') {
+        // TODO: implement what happens with functions (helper?)
+        defineProperty(prototype, name, descriptor);
+      }
+
+      return prototype;
+    }, prototype);
+
+  }
+
+  function defineTransitions(prototype, transitions) {
+    // property is transitions hash
+    transitions = assign({}, transitions, {
+      replace(current, ...args) {
+        return new this.constructor(...args);
+      }
+    });
+
+    getOwnPropertyNames(transitions).forEach(function(name){
+      let descriptor = getOwnPropertyDescriptor(transitions, name);
+      defineProperty(prototype, name, makeTransitionDescriptor(descriptor.value));
+    });
+  }
+
+  function makeTransitionDescriptor(transition) {
+    return {
+      value: function() {
+        let result = transition.call(this, this.value, ...arguments);
+        if (result instanceof this.constructor){
+        // transition returned new microstate - use it
+        return result;
+        } else if (typeof result === 'object') {
+        // transition returned an object - merge it
+        let next = new this.constructor();
+        return Object.assign(next, this, result);
+        } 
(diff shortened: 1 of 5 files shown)


taras:
This [handbook](https://github.com/thejameskyle/babel-handbook/blob/master/translations/en/plugin-handbook.md) is very useful to understanding how to build a Babel plugin. Requiring a babel plugin can be a barrier to adoption. This is something that we should treat with caution.

There might be other ways to reduce complexity. We should look into those first.

cowboyd:
> It takes substantial cognitive effort to keep track of what's happening when a class is defined vs in runtime when transitioning state. Any effort that we put towards making it clear what's executed in design time vs run time will go a long way towards making the code easier to understand and evolve.

Agree, although I think that most of this will need to education around what the `transitions` hash actually is, and that it will end up as a method on the instance that will _always return a new instance of the same microstate class_ no matter what, even if you don't return an instance from it. I think that is the lynch pin of the library and so we just have to hammer the point home in the README.

> Having constructors that transform user input to properties that are merged into the object complicated implementation considerably. It also makes it difficult to reason about how data should be formatted to be merged into state. I found it a lot easier to have no constructors for nested state. We might want to eliminate constructors to keep things simple.

This makes a lot of sense. It occurs to me that when we declare the type of the content, we are _effectively declaring a constructor_ for you, and so really, it seems like only in the very advanced cases (like declaring primitives) should users ever write a initialize transition.

Basically, I think that the only true constructor should be on the base `Microstate` object itself, and all it does is accept an object of attributes to assign to itself along with a list of overrides (to support ease of merging). Subclasses just declare their "structure" and then the actual constructor is generated for you. 

For example, the Microstate constructor could look like this:

```javascript
class Microstate {
  constructor(current, attrs) {
    Object.assign(this, current, attrs);
  }
}

// alternative one
export default Microstate({name: string, todos: list(TodoItem)}).extend({
});

// alternative two
export default Micostate.extend({
  structure: {
    name: string,
    todos: list(TodoItem)
  }
})
```

> We need a clear way to know what needs to be merged. I tried deducing it from target object, but it's difficult to know because some properties are on the prototype and others on the object. Prototype chain complicates this further because we need to accumulate all of the properties from the prototype chain.

The properties that need to be merged are those that have previously been merged, as in those that would be assigned via `Object.assign`. I may be missing something here, so I'd probably have to look at the code.

taras:
> the lynch pin
Just to clarify, the cognitive overhead is on the developer writing Microstates library (i.e. me and you 😄 ) With the right documentation, the idea of returning an instance is a clear rule and it'll be easy to explain to people. So we're all good there (I think). 

> `structure` property
I like the `structure` property in concept. We need that level of clarity, but maybe not that specific solution. 

> I may be missing something here, so I'd probably have to look at the code.
This becomes a lot easier if we say that we only merge microstates. This was tripping me up because some of our tests have properties on Microstate definition. We could allow for values to be passed into a declaration if we automatically turn it into a micro state.

```js
WorkTodoList = TodoList.extend('WorkTodoList', {
          constructor(name = '', items = [])  {
            return {name, items};
          },

          get upperName() {
            return `☭ ${this.name.toUpperCase()}`;
          },

          due: 'today',

          transitions: {
            changeDue(current, due) {
              return {
                due: due
              }
            }
          }
        });
```

We could convert `due` to a string micro state on the class declaration.

src/utils/prototype-for.js
```diff
@@ -0,0 +1,146 @@
+import MicroState from '../microstate';
+
+const { getOwnPropertyNames, getOwnPropertyDescriptor, defineProperty } = Object;
+
+export default function prototypeFor(prototype, attrs) {
```
cowboyd: I think the way to avoid having a hidden property is to make `prototypeFor` itself a recursive function that takes a context object, and that every time you recurse into a nested substate, you push that substate onto the context.

Basically, each substate will serve as the prototype for the nested substate's prototype :)

cowboyd: A perhaps different way of saying it is that I think the key thing that we're fighting is that I think we want to generate a prototype per instance (which has the class prototype as _it's_ prototype, but with it's own overrides to represent the properties it uniquely holds) as opposed to a single class prototype that handles all cases.

taras:
Bye bye issue...

PR #12, opened 2016-08-26
Simple Reboot. Simple State.

cowboyd:
In order to make forward progress, this brings back the absolute simplest thing that could possibly work: a bare-bones node/mocha test harness that can be iterated on quickly.

For reasons of debuggability and cross browser testability, we will need to sort out how to get these tests running in the browser at some point, but we can do that in parallel with the actual work on the microstate library.

Included with this change is the basic `State` object that wraps any JavaScript object value. It contains the four basic operations common to all microstates `set`/`assign`/`put`/`delete`. In the end, I decided not to make a separate branch of microstate for primitive types such as `number`, `boolean` etc... since they _can_ wrap primitive values.

Note that these operations are not currently implemented as transitions. Instead, they are just simple methods. The plan is to make the state object dogfood the transition mechanism for `set`, `assign`, `put` and `delete`.


taras:
Awesome start!!! 👍

PR #14, opened 2016-09-06
Make microstates composable

cowboyd:
This adds the ability to nest microstates within other microstates, and to invoke transitions on nested microstates that will return a new version of the entire datastructure.

```js

let state = new State({
  one: new State({
    two: new State({
      three: new State(3)
    })
  })
});
```

The `valueOf()` method for the super-state will return a recursively unboxed value of all the states:

```js
state.valueOf() //=> {one: {two: {three: 3}}}
```

State transitions can be invoked at any level of the tree, but will result in a global state transition:

```js
let next = state.one.two.three.set('three')
next === state //=> false
next.valueOf() //= {one: {two: {three: 'three'}}}
```

But the original state remains unchanged:

```js
state.valueOf() //=> {one: {two: {three: 3}}}
```

### DIFF
--- a/src/computed-property.js
+++ b/src/computed-property.js
@@ -1,3 +1,4 @@
+import assign from './assign';
 /**
  * Property Descriptor that computes its value once and then permanently caches
  * the result.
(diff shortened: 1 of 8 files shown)


taras:
😱 Yess!

taras:
Is there anything missing before we can merge this?

cowboyd:
I don't think so. I didn't get to start a README, which I'd very much like to get going on, but I don't think there's anything holding up.

PR #19, opened 2016-09-29
Overloadable value of

taras:
This pull request adds the ability to overload the valueOf method on a microstate class. The valueOf method receives boxed in value of the microstate and returns that value for the default implementation.

```js
let MyState = State.extend({
  happy: new State(true)
  valueOf(value) {
    return {
      feeling: value.happy ? "😀" : "😧"
    }
  }
})
```

This PR includes 2 failing tests:
- access to prototype values in valueOf
- inherited custom valueOf


taras:
@cowboyd this is ready for review and merge

PR #20, opened 2017-08-20
Classy microstates

taras:
This PR introduces the API that we discussed a month ago. It's using ES6 classes to describe the structure of a microstate and uses these classes to automatically build transitions. It includes many of the wonderful things that we wanted:

1. State is serialized to POJO and deserialized from POJO
2. It uses Haskell lenses via ramda.js to allow transitioning composed Microstates and always retrieving  new objects on transition
3. It supports computed properties that are evaluated on composed microstates
4. It provides Observable interface to receive state transitions
5. It supports return new store/actions on transition interface for more imperative style programming
6. It allows composed microstates to be custom classes to allow users to customize their composted microstates
7. Custom transitions are static methods and prevent access to mutating the state
8. It's written in typescript
9. Uses ComputedProperty to cache and lazily compute composed states
10. Supports primitive values without wrapping or valueOf
11. Actions are now stable. They do not change when state changes. This will prevent unnecessary re-renders in React and Glimmer
12. Added the ability to compose arrays using `[Product]` syntax
13. supports valueOf()
14. Supports merge transition on composed states
15. Supports set transition on composed states and parameterized arrays
16. `new TypeTree(Type)` builds schema tree from type

```js
import microstates from 'microstates';

let { state: count, transitions } = microstates(Number);

console.log(count);
//=> 0

console.log(transitions.increment())
// 1

console.log(count);
//=> 0
```

You can restore state by proving data as the second argument.

```js
import microstates from 'microstates';

// class name here doesn't matter
class State {
  user = User
}

class User {
  firstName = String;
  lastName = String;
  get fullName() {
     return `${this.firstName} ${this.lastName}`;
  }
}

let { state } = microstates(State, { user: { firstName: 'Peter', lastName: 'Griffin' } });

console.log(state.fullName);
//=> Peter Griffin
```

To subscribe to the state,

```js
import {toObservable} from 'microstates';

let ms = toObservable(Number);
let { transitions } = ms;

let subscription = ms.subscribe({
  next(newState) => console.log(newState));
});
//=> 0

console.log(transitions.increment());
//=> undefined
//=> 1
```

To use with React, you would do something like this,

```js
import React from 'react'
import ReactDOM from 'react-dom'
import { toObservable } from 'microstates';

const rootEl = document.getElementById('root')

let ms = toObservable(Number);

let { 
  transitions: { increment, decrement }
} = ms;

const render = () => ReactDOM.render(
  <Counter
    value={count}
    onIncrement={increment}
    onDecrement={decrement}
  />,
  rootEl
)

render();

let subscription = ms.subscribe({
  next(newState) => {
   count = newState;
   render();
  }
})
```

~I haven't tried this yet, but in theory, it should work like this.~

## Array Composition

It's now possible to compose arrays by using parameterized array syntax. This part is a little bit shady but it'll give us enough to talk about. 

```js
class Product {
  title = String;
}

class State {
  products = [Product]
}

let { state, transitions } = microstates(State, { products: [ title: 'MacBook' ] });

let newState = transitions.products[0].title.concat(' Pro');

console.log(newState);
// { products: [Getter] }

console.log(newState.products[0].title);
// MacBook Pro
```

## set transition

Set transition makes it possible remove a branch of a tree of state or replace it with another branch of same type. 

```js
class Baz {
  name = String;
}

class Foo {
  name = String;
  baz = Baz;
}

let ms = microstates(
  class State {
    foo = Foo;
  }
);

let sub = microstates(Foo, { name: 'Foo', baz: { name: 'Baz' } });

console.log(ms.transitions.foo.set(sub));
// => { name: 'Foo', baz: { name: 'Baz' } }
```

## state.valueOf()

Composed states and parameterized arrays have valueOf method that returns boxed in value. `set` requires that passed in value is of same type as the type that you're trying to set. The easiest way to generate a data structure of same type is to create a new microstate using type that you'll be setting.

```js
let { state } = microstates([String]);
console.log(state.valueOf());
//=> undefined
```

```js
class User {
  name = String;
}

class Authentication {
  isAuthenticated = Boolean;
  user = User;
}

class State {
  authentication = Authentication;
}

let { state } = microstates(State, { authentication: { user: { name: 'Peter' } } });

console.log(state.authentication.valueOf());
//=> { user: { name: 'Peter' } }

console.log(state.authentication.user.valueOf());
//=> { name: 'Peter' }

console.log(state.authentication.isAuthenticated);
//=> false
```

## merge transition

Merge transition will merge new state into existing state. Merge follows same rule as `set` transition which is that it requires the passed in value to be of same type as value that you're trying to set.

```js
 class Product {
  name = String;
}

class User {
  name = String;
}

class Session {
  token = String;
}

class Authentication {
  isAuthenticated = Boolean;
  session = Session;
}

class State {
  products = [Product];
  authentication = Authentication;
  currentUser = User;
}

initial = microstates(State, {
  currentUser: { name: 'Anonymous' },
  products: [{ name: 'MacBook' }],
});

let authentication = microstates(State, {
  authentication: { isAuthenticated: true, session: { token: 'ABC' } },
  currentUser: { name: 'Peter Griffin' },
});

let merged = initial.transitions.merge(authentication.state);

console.log(merged);
//=> { 
//         products: [ { name: 'MacBook' } ],
//         authentication: { isAuthenticated: true, session: { token: 'ABC' } },
//         currentUser: { name: 'Peter Griffin' }
//      }

```

## TODO

- [x] Implement set on composed states
- [x] Merge action
- [x] Refactor for action to return state
- [x] Add Array composition - ramda lenses support it
- [x] Make result of `microstates` call observable

## Future (not this PR)

1. Implement Queries
2. Polymorphic types
3. Add common actions to primitive states
4. Implement async type
5. Lots of testing and demos

But I think we got it :)


taras:
> I really like this way of defining the schemas

I'm glad :)

> why not use instance methods instead of static function properties?
> reads a lot more user-friendly than(static)

I agree completely, but their certain purity and logic to using static. Static methods do not have access to the instance. This is symmetric to how actions work because they don't have access to the instance. If we make actions instance methods, then we'll loose the ability to differentiate between actual instance methods and actions. 

We will lose the ability to do something like this:

```js
class List {
  filter(...args) {
    return this.filter(...args);
  }
}
```

This is a contrived example, but I hope it shows that there might be use cases where an instance method would operate on the current object without transitioning it. 

> polymorphism - what if I have a list of records, but the records could be of different type? how do I represent that .How do I reconstitute that when "restoring" a microstate

What about this?

```
class WidgetContainer {
  widgets = List.of([Wrench, Screwdriver, Drill], serialized => ( 
    switch(serialized.type):
      case 'wrench':
        return Wrench;
     case 'screwdriver':
        return Screwdriver;
     case 'Drill':
        return Drill;
     default:
        return Item;
    ));
}
```

> parametric types. what if part of my structure is

I'm not sure. I'm open to discussion. I didn't have a chance to think about List very much yet. We have some bits to discuss here.

> I'm loving the code

😄 

> the tests although especially in the wrapping parts, it could use a lot more documentation to help determine what's going on and why.

Yes, I'm going to clean this part up and write tests for it. I'm going to simplify it a lot.

src/microstates.ts
```diff
@@ -0,0 +1,44 @@
...
+    observer = _observer;
+    return {
+      unsubscribe,
+    };
+  };
+
+  let onChange = (newState: IState) => {
+    if (observer) {
+      observer.next(microstate(Class, newState));
+    } else {
```
taras: > In other words, the transitions aren't pure functions, but instead depend on some shared mutable state. This comes from intermingling the logical transitions of a state (pure) with the tracking of that state over time (impure).

Yes.

> In other words, a microstate is data + transitions, and an observable hold the microstate and tracks transitions.

Yes!, but it'd need to be something like this:

```js
let ms = microstate(Boolean, false);

let observable = Observable.from(ms);

let { state, actions } = ms;
```

To do this, our `{ state, actions }` needs to have [`{}[Symbol.observable}`](https://github.com/tc39/proposal-observable#observablefrom) on it. This will allow ES7 Observable convert our hash into an observable.

taras:
@cowboyd I refactored the utilities and gave them better names. I'll work on tests for the utils.

taras:
I'm going to work on the arrays more today. I'll add more tests because current implementation is very flaky.

taras:
@cowboyd refactored the code to introduce `TypeTree` class. `new TypeTree(Type)` will create a handy type tree that is easy to map. Each node has properties that describe it. I added `mapState` & `mapTransitions` functions that accept a `tree` and callback that'll be lazily executed for each node.

I eliminated the word actions from the repo (I think). To perform transitions, you must now call `let { state, transitions } = microstates(Type);`.

I added `toObservable` function which takes a `Type` and returns an instance of `MicrostateObservable`. `new MicrostateObservable` has `transitions` object that'll push new states to the observer.

I'd like to merge this PR. Thoughts?

PR #21, opened 2017-09-05
WIP: Adding polymorphism

taras:
I'm starting to work on polymorphism. I want to get the conversation going about how this should work. Here are my early thoughts. What do you think?

### DIFF
--- a/src/Interfaces.ts
+++ b/src/Interfaces.ts
@@ -62,5 +62,3 @@ export interface ITypeTree {
 export interface ITypeTreeProperties {
   [name: string]: ITypeTree;
 }
-
-export interface IPolymorphic<T> {}
--- a/src/Interfaces.ts
+++ b/src/Interfaces.ts
@@ -62,3 +62,5 @@ export interface ITypeTree {
 export interface ITypeTreeProperties {
   [name: string]: ITypeTree;
 }
+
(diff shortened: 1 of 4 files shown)


cowboyd:
I think general parameterization of classes needs to come first, and then the polymorphism can depend on that.

```

import { Parameterized } from 'microstates';

//note that Parameterized can be made a decorator later.
class List extends Parameterized(Object) {
//...  
}

microstate(List.of(Dog))
microstate(List.of(Record))
```
With this in place, then you don't need any special casing for arrays. `[Type]` is just a macro for `ArrayPrimitive.of(Type)` and the two syntaxes could be used interchangeably.

If we can get our array primitive to dogfood this mechanism, then we'll have it everywhere else and it should take care of most cases where polymorphism would normally be reached for. I'd think you want to only use it as a last resort where you can't know the type information ahead of time.

cowboyd:
Unfortunately, I don't see a way around some form of algebraic data types :(

```
class ArrayPrimitive extends Parameterized(or(Object, Number, String, Boolean)) {
}
```

That  doesn't let you have booleans in the same array as numbers, but what it allows you to do is have `ArrayPrimitive.of(Object)` and `ArrayPrimitive.of(Boolean)` to both be valid classes. so that I can have [{}] and [false] both be valid instances

PR #23, opened 2017-09-06
Simplified getValue by removing initializer

taras:
* getValue no longer received an initializer as the first argument
* initializer is called explicitly in mapState
* initializer no longer expects `(current, newState) => newState`
* simplified initializer to just return initialized value


taras:
@cowboyd ping

PR #24, opened 2017-09-25
Functor, tree mapping and reduced feature set

taras:
This PR adds tree mapping that is inspired by Functors. It's not using `funcadelic.js` but implements many of the ideas introduced there. 

The biggest improvements are in the following files:

1. [Tree](https://github.com/cowboyd/microstates.js/blob/functor/src/utils/tree.ts) - primitive for mapping Type Trees to Mappable Trees
2. [Transitions](https://github.com/cowboyd/microstates.js/blob/functor/src/utils/transitions.ts) - provides map function for lazily mapping transitions of a Type Tree
3. [State](https://github.com/cowboyd/microstates.js/blob/functor/src/utils/state.ts) - provides map function for lazily mapping state and `from` function for turning Type Tree and value to state.
4. [Microstates](https://github.com/cowboyd/microstates.js/blob/functor/src/microstates.ts) - Converts a Type Tree and value into a states tree and transitions tree. 

Removed in this PR:
1. Any mention of Parameterized Arrays and `[<Type>]` syntax.

Changes in this PR:
1. change `microstates(Type)` to `Microstates.from(Type)`
2. moved to instant methods for transition definition instead of static methods

# TODO
- [x] Make Tree a Functor
- [x] Change `State.from(tree)` to use `map` from `funcadelic`
- [x] Change `Transitions` to be a functor
- [ ] Refactor use Cast TypeClass (when it's finalized)
- [ ] Make Transitions a TypeClass for transitions
- [ ] Make Microstates a TypeClass for initialize


src/primitives/boolean.js
```diff
@@ -0,0 +1,8 @@
+export default class MicrostateBoolean {
```
cowboyd: Let's call this `BooleanState` since `Microstate` is the less significant thing about them and `Boolean` is the more significant thing. Same applies for the other primitives.

src/utils/get-own-property-descriptors.js
```diff
@@ -0,0 +1,7 @@
+export default function getOwnPropertyDescriptors(o) {
+  return Object.getOwnPropertyNames(o).reduce((accumulator, propName) => {
```
cowboyd: How is this different from `Object.getOwnPropertyDescriptors`?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptors

src/utils/tree.js
```diff
@@ -0,0 +1,108 @@
...
+      data: {
+        get() {
+          return {
+            name,
+            path,
+            type: getReducerType(Type),
+            schemaType: Type,
+            isPrimitive: primitive,
+            isComposed: !primitive,
+            isList: list,
```
cowboyd: Where is this information used?

src/utils/state.js
```diff
@@ -0,0 +1,39 @@
...
+   * a callback and a tree to map. The callback is invoked lazily when value
+   * is read. The callback will receive a view function as first argument.
+   *
+   * The view function has curries lensPath(path). Invoking the view with value,
+   * will return the focused value.
+   *
+   * @param fn (view: (state: any) => any, path: [string | int]) => any
+   * @param tree Tree
+   */
+  static map(fn, tree) {
```
cowboyd: Where is this method used outside of `State.from`?

If we need a map function for state, then maybe have the State constructor take a tree, and then implement its own instance of `Functor` that maps over that tree.

src/utils/transitions-for.js
```diff
@@ -0,0 +1,40 @@
...
+        [name]: name === 'initialize' ? descriptor.value : transition(descriptor.value),
+      });
+    },
+    {
+      set: transition(function set(current, state) {
+        return state && state.valueOf ? state.valueOf() : state;
+      }),
+    }
+  );
+  let isComposed = type === Type;
```
cowboyd: Is this necessary? If we've played our cards right, then the entire recursive structure should be captured in the `Functor` instance for tree. If we have to worry about that as we gather information about a node, then we've lost.

src/utils/tree.js
```diff
@@ -0,0 +1,108 @@
...
+          );
+        },
+      },
+    });
+  }
+}
+
+Foldable.instance(TypeTree, {
+  foldl(fn, initial = {}, tree) {
+    return foldl((memo, args) => fn(memo, args), initial, tree.children);
```
cowboyd: This folds the children, but not the actual node, so I don't think this is a valid fold.

src/utils/tree.js
```diff
@@ -0,0 +1,108 @@
...
+
+Functor.instance(TypeTree, {
+  /**
+   * Lazily invoke callback on every proprerty of given tree,
+   * the return value is assigned to property value.
+   * 
+   * @param {*} fn (TypeTree, path) => any
+   * @param {*} tree Tree
+   */
+  map(fn, tree) {
```
cowboyd: By the same token, this is not a valid Functor instance. `map` _must_ return a `TypeTree`

Here is how I would implement map:

```js
  map(fn, tree) {
    return Object.create(TypeTree, {
      path: {
        get() { return tree.path; }
      },
      // map this node.
      data: {
        get() { return fn(tree.data); }
      },
      // map the children.
      children: {
        get() {
          return map(child => map(fn, child), this.children);
        }
      }
    });
```
 

I'd go ahead and make `path` a first-class attribute of the Tree node. It won't ever change when you map. The only thing that can change is the `data` inside the tree node.

src/utils/get-own-property-descriptors.js
taras: It's not but I couldn't get it to work in typescript. I can use a polyfill now.

src/primitives/boolean.js
taras: Sounds good

src/index.js
```diff
@@ -0,0 +1,33 @@
...
+      transition => (...args) => transition(this.states, ...args),
+      Transitions.from(tree)
+    );
+  }
+  /**
+     * Create new Microstates for same type tree but new value.
+     * @param value any
+     */
+  to(value) {
+    return new Microstates(this.tree, value);
```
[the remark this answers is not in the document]
taras: Yes, this is doable.

src/utils/tree.js
```diff
@@ -0,0 +1,94 @@
...
+       *
+       * class Person {
+       *  parent = Person;
+       * }
+       *
+       * Eager evaluation would cause infinite loop because parent would be evaluated recursively.
+       */
+      children: {
+        get() {
+          return reduceObject(
```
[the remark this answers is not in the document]
taras: Good idea

src/utils/tree.js
taras: It might not be anymore. It was used with parameterized array

src/utils/transitions-for.js
taras: I'll look into this

src/utils/tree.js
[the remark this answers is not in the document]
cowboyd: Actually, I like the way you had it better. It makes the `tree.collapse` trivial to implement with nothing but `append` + `map` instead of needing a fold.

PR #26, opened 2017-10-21
Adding some initial README

taras:
Add some initial documentation.


README.md
```diff
@@ -1,26 +1,643 @@
...
+the structure of your state and `value` that is the initial state.
+
+```js
+// initial value is undefined so state will be default value of String which is an empty string
+microstate(MS.String).state;
+// => ''
+
+// initial value is 'hello world' so state will be hello world.
+microstate(MS.String, 'hello world').state;
+// => ''
```
sivakumar-kailasam: needs to be 'Hello world' here.

README.md
```diff
@@ -1,26 +1,643 @@
...
+necessary. Microstates will take care of ensuring that all transitioning composed microstates is
+done immutably.
+
+Defining custom types is done using ES2016 classes syntax. In JavaScript classes are functions. So
+any function can work as a type, but we recommend using class syntax.
+
+```js
+microstate(class {});
+// => {
+//      merge: Function
```
sivakumar-kailasam: I don't see merge function when I tried this in repl. Is this a WIP?

taras: No, this is a mistake in the documentation. Merge function will only appear when a type has nested states.

PR #30, opened 2017-10-21
Eliminated mapping of native types to built in types

taras:
This PR eliminates internal resolution of native types to built-in types. This change should make the ecosystem of types more consistent by eliminating hardcoded types. It'll be possible to change native type by changing what's in-scope of definition module.

### Changes

* isPrimitive - now returns true only when a type has composed types.
* microstates now exports `import { Number, String, Boolean, Array, Object as ObjectType } from 'microstates'`
* removed getType

*Note*:  babel uses a function that calls a function on `Object`, when I import Object it overwrites the default Object and throws an error.


taras:
I'm using `import { Object as ObjectType } from 'microstates'`, should we use `import ObjectType from 'microstates'` instead?

PR #33, opened 2017-10-27
Transition context and batching transitions

taras:
Like #20 and #25, this PR turned into a bit of a runaway train. 

The original goal of this PR was to add transition context and the ability to batch transitions. I'll items everything that this PR introduces and how it can be used.

## `microstate(Type, value)`

Previously, to create a Microstate we exposed `Microstates.from(Type, value)`. This has now been replaced with `microstate(Type, value)`.

## `microstates(Type, value).transitions` is now hidden

Previously, `let { transitions } = Microstates(Type, value)` provided a collapsed object for mapping transitions. Now the transitions are hidden on result of `Microstates(Type, value)`. To retrieve metadata use the following,

```js
import { reveal } from 'microstates/src/utils/secret';
import microstate, * as MS from 'microstates';

let ms = microstate(MS.Number);
ms.transitions
// => undefined

reveal(ms);
// => { Type: MS.Number, value: undefined, state: Tree, transitions: Tree }
```

This should only be necessary for integrations.

## Chainable transitions

It was not possible to chain transitions because transitions returned the value which needed to be wrapped in Microstates. Closest was the `.to` method, but it was a weird API. Now, root level transitions are automatically chainable.

```js
microstate(MS.Number).increment().increment().increment().valueOf();
// 3
```

## Transition Context

This was the original purpose of this PR. It's to allow patching transitions inside of a transition. In our earlier conversations, we discussed `this(current)` syntax. This has been changed to support parity with Microstates constructor signature. `this()` is equivale to `this(Type, value)` but `Type` and `value` are provided by default. You can override the default `Type` and `value` with `this(AnotherType, anotherValue)`.

```js
class ModalContent {
  text = String;
}
class Modal {
  isOpen = Boolean;
  title = String;
  content = ModalContent;
}
class State {
  messages = Array;
  modal = Modal;

  addItemAndShowModal(current, message, prompt) {
    return this()
      .messages.push(message)
      .modal.isOpen.set(true)
      .modal.content.text.set(prompt);
  }
}
```

## Constants Support

Constants make it possible to define properties on Types that can't be transitioned. These values are meant to be used with Type-shifting to create nodes that transition through multiple states.

```js
let ms = microstate(
  class Type {
    n = 10;
    b = true;
    s = 'hello';
    o = { hello: 'world' };
    a = ['a', 'b', 'c'];
    null = null;
    greeting = MS.String;
  }
);

ms.state.n
//=> 10

ms.state.null
//=> null

ms.state.b
//=> true

ms.valueOf();
// undefined
```

Static values are shared by all instances of the Type.

```js
microstate(Type).state.o === microstate(Type).state.o;
```

## Type-shifting

This is what I called being able to change the structure of the Type tree at runtime. An important quality of Type-shifting is that the transitioned types are intended to be an intermittent and discardable state. It makes it possible to represent state machines like File Uploaders. Type-shifting is done using the transition context.

```js
class Line {
  a = MS.Number;
  add({ a }, b) {
    return this(Corner, { a, b });
  }
}

class Corner extends Line {
  a = MS.Number;
  b = MS.Number;
  add({ a, b }, c) {
    return this(Triangle, { a, b, c });
  }
}

class Triangle extends Corner {
  c = MS.Number;
}

```

^^ see this example in tests https://github.com/cowboyd/microstates.js/blob/c1bfdc98afb625a49e0945ac9b812c26506d582a/tests/microstates.test.js#L591

## Type-shifting + Constant values

Together these two features make it possible to model state machines that have multiple related properties. 

```js
class Async {
  content = null;
  isLoaded = false;
  isLoading = false;
  isError = false;

  loading() {
    return this(AsyncLoading);
  }
}

class AsyncError extends Async {
  isError = true;
  isLoading = false;
  isLoaded = true;
}

class AsyncLoading extends Async {
  isLoading = true;

  loaded(current, content) {
    return this(
      class extends AsyncLoaded {
        content = content;
      }
    );
  }

  error(current, msg) {
    return this(
      class extends AsyncError {
        error = msg;
      }
    );
  }
}

class AsyncLoaded extends Async {
  isLoaded = true;
  isLoading = false;
  isError = false;
}
```

^^ see this example in tests https://github.com/cowboyd/microstates.js/blob/c1bfdc98afb625a49e0945ac9b812c26506d582a/tests/microstates.test.js#L642


src/microstates.js
```diff
@@ -1,28 +1,30 @@
...
+const { assign } = Object;
+
+export default function Microstates(Type) {
+  let tree = Tree.from(Type);
+
+  return function Microstate(value) {
+    let states = States(tree, value).collapsed;
+    let transitions = Transitions(tree, states, value);
+
+    return assign(
```
cowboyd: Curious why `assign` here and not `append`. This will eagerly evaluate the object that you're mapping with line 26

src/microstates.js
```diff
@@ -1,28 +1,30 @@
...
+      {
+        Type,
+        states,
+        transitions: transitions.collapsed,
+
+        valueOf() {
+          return value;
+        },
+      },
+      map(transitions => map(t => (...args) => Microstate(t(...args)), transitions), transitions)
```
cowboyd: This is definitely making my head spin. What is the goal here?

src/microstates.js
taras: They would both work. In this case, it seems that `assign` was the least powerful tool for the job.

src/microstates.js
taras: This makes it possible to chain transitions for both context and microstates.

For example,

```js
let NumberMicrostate = Microstates(Number);
let n = NumberMicrostate(10);
n.increment().increment().increment().valueOf();
// => 13
```

src/microstates.js
taras: I originally used object spread but it's not available before Node 8.

cowboyd:
I've had the opportunity to use the new API and it “feels” really good. There are definitely a lot of changes and improvements I can see that can be made to the internals, but if the public API is right, then that we can roll with it now, and refactor later.

I would like to propose the following tweaks:

- [ ] The convention for the primary function is `microstate` singular. E.g.

```js
import microstate, { Boolean } from 'microstates';

let bool = microstate(Boolean).toggle().toggle().state // singular
```

- [ ] `.states` property becomes `.state`
- [ ] `.value` is not a property, just call `valueOf()` if you need it.
- [ ] valueOf() is not an enumerable property (this is standard JavaScript behavior)
- [ ] the `states` property, just becomes `state`
- [ ] it is an error to name a transition, or property `state`, or `Type`
- [ ] the `transitions` property, (if needed, is it?) is hidden somewhere.
- [ ] the `microstate` property, (if needed, is it?) is hidden somewhere (behind metadata accessed via symbol?)
- [ ] fix #40 and #41 
- [ ] squash commits.

With these API changes tested and in place, it's an API that we can start to really lean on, so let's leave the internals as-is. Time to ship shine.


[1:59] 
we’re so close to the public api, that it’s time to ship and shine.

taras:
@cowboyd pls review #42 and #43. The only thing left to do is squash, which I'll do when I hit "Squash and Merge"

PR #43, opened 2017-12-06
Make static properties fully static

### LINKED ISSUE #41: Static properties for a microstate class are not static
If a property is inherent to a microstate class, then all instance ought to be able to share it.

```javascript
import Microstates, * as MS from 'microstates';


let ms = Microstates(MS.Boolean, true);
console.log("ms.toggle().valueOf() = ", ms.toggle().toggle().toggle().valueOf());

class Confirmation {
  constructor() {
    this.unique = {};
  }
  get isArmed() { return false; }
  get isConfirmed() { return false; }

  arm() {
    return this(Armed);
  }
  reset() {
    return this(Confirmation);
  }
}

class Armed extends Confirmation {
  get isArmed() { return false; }

  confirm() {
    return this(Confirmed);
  }
}

class Confirmed extends Confirmation {
  get isConfirmed() { return true; }
}

let dnd = Microstates(Confirmation);
console.log('initial', dnd.states.isArmed);

let result = dnd
    .arm()
    .confirm();

let result2 = dnd
    .arm()
    .confirm();

console.log('unique 1', result.states.unique);
console.log('unique 2', result2.states.unique);
console.log('result.states.unique === result2.states.unique', result.states.unique === result2.states.unique);
```

outputs

```
ms.toggle().valueOf() =  false
initial false
unique 1 {}
unique 2 {}
result.states.unique === result2.states.unique false
```

taras:
This PR makes sure that once a type is instantiated, all following instances of that Type reference same static values as the first instance.

Closes #41

### DIFF
--- a/src/utils/secret.js
+++ b/src/utils/secret.js
@@ -1,4 +1,6 @@
-const symbol = Symbol('🙈 microstate');
+const symbol = Symbol('🙈');
+
+const { getOwnPropertySymbols } = Object;
 
 export function keep(object, value) {
   return Object.defineProperty(object, symbol, {
@@ -8,9 +10,7 @@ export function keep(object, value) {
 }
 
 export function reveal(object) {
-  if (object) {
+  if (object && getOwnPropertySymbols(object).includes(symbol)) {
     return object[symbol];
-  } else {
-    return object;
   }
 }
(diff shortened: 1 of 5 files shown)


src/utils/secret.js
```diff
@@ -1,4 +1,6 @@
-const symbol = Symbol('🙈 microstate');
+const symbol = Symbol('🙈');
```
cowboyd: 😂 😂 😹

PR #45, opened 2017-12-07
Make it possible to type shift from a type constructor

taras:
Previously transitions were built from Type ignoring the initialized state. This meant that state and transitions tree structure were mismatched. This PR generates transitions from Type of initialized state. 

I moved some code around to make them into their own functions.


taras:
@cowboyd ping

PR #51, opened 2017-12-14
Added Microstate Functor for mapping transitions

taras:
Microstate objects are nested transitions with a state property. Mapping microstate object is effectively mapping transitions. To map state one can do `map(v => v, ms.state)`. 

This PR adds Functor instance for Microstate class to allow mapping microstate transitions.


taras:
@cowboyd ping

Issue #49, opened 2017-12-12
Value of getters should not be added to valueOf after custom transition


taras:
Expect: custom transitions should not add the value of getters to `valueOf()`

Actual: getters are materialized and added to `valueOf`

To reproduce: failing test in https://github.com/cowboyd/microstates.js/commit/f1a83177bdaf9de594d9aa2147efdae3cc76b4d4

taras:
Resolved in #master

PR #53, opened 2017-12-17
noop transitions should do nothing

taras:
This PR has commits from #52 and addresses #49 by making sure that custom transitions only receive a slice of value rather than slice of state.

### DIFF
--- a/src/utils/state.js
+++ b/src/utils/state.js
@@ -35,6 +35,10 @@ export default function state(root, value) {
 
           let val = t.call(context, current, ...args);
 
+          if (current === val) {
+            return microstate(root, value);
+          }
+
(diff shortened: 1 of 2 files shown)


sivakumar-kailasam:
Looks like your prettier config removed trailing comma from many places causing a huge diff. Can you fix that? I can raise a separate PR to run prettier via husky & lint staged.

taras:
@cowboyd ping

cowboyd:
what if I want to use some of the computed properties from state? But now my transition is only getting passed the value.

taras:
>  But now my transition is only getting passed the value.

I don't see this. Microstate is instantiated from the slice of the value, so you're still getting a full state with computed properties.

PR #55, opened 2018-01-14
Do not loose root type on custom transitions

taras:
The root type should remain unchanged when a custom transition did not change the type of node.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
   },
   "dependencies": {
     "@std/esm": "^0.14.0",
-    "funcadelic": "^0.2.6",
+    "funcadelic": "0.2.6",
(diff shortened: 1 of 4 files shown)


cowboyd:
Although looks like there's still a failing test.

package.json
```diff
@@ -43,7 +43,7 @@
   },
   "dependencies": {
     "@std/esm": "^0.14.0",
-    "funcadelic": "^0.2.6",
+    "funcadelic": "0.2.6",
```
cowboyd: Curious why.

taras: It's sliding to 0.2.7. I'll need to investigate why 0.2.7 is breaking tests.

PR #61, opened 2018-01-19
Setting up Functor instances from index.js

taras:
For Functor map to work correctly Functor.instance needs to be setup before `map` is used anywhere. This PR moves typeclasses setup to index.js module to ensure that instances are setup before anything from microstate library can be used.

### DIFF
--- a/src/typeclasses.js
+++ b/src/typeclasses.js
@@ -1,0 +1,32 @@
+import { Functor, map } from 'funcadelic';
+import { Microstate } from './microstate';
+import { reveal } from './utils/secret';
+import Tree from './utils/tree';
+
+Functor.instance(Microstate, {
+  map(fn, microstate) {
+    let { transitions } = reveal(microstate);
+    return map(transitions => map(transition => fn(transition), transitions), transitions)
+      .collapsed;
+  },
+});
+
+Functor.instance(Tree, {
+  /**
+   * Lazily invoke callback on every property of given tree,
+   * the return value is assigned to property value.
+   *
+   * @param {*} fn (TypeTree, path) => any
+   * @param {*} tree Tree
+   */
+  map(fn, tree) {
+    return new Tree({
+      data() {
+        return fn(tree.data);
+      },
+      children() {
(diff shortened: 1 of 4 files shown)


src/typeclasses.js
```diff
@@ -0,0 +1,32 @@
+import { Functor, map } from 'funcadelic';
+import { Microstate } from './microstate';
+import { reveal } from './utils/secret';
+import Tree from './utils/tree';
+
+Functor.instance(Microstate, {
+  map(fn, microstate) {
+    let { transitions } = reveal(microstate);
+    return map(transitions => map(transition => fn(transition), transitions), transitions)
```
cowboyd: I just noticed that this does not appear to be a valid functor. The type of `map` for a Functor `F` has to be `F a -> (a -> b) -> F b`

Which basically says, the _exact_ structure must remain in tact, not just an analogous structure. So you give it a function `a -> b` and it uses it to turn `F a` into `F b` but the `F` remains the same.

This is a long way of saying that for `Microstate` to properly inhabit functor, its `map` method must return a `Microstate`. 

This has me thinking that perhaps we're missing a separate utility Functor. like a a `CollapsedTree` or the `Structure` or what have you. Because, for a given microstate, the transitions tree, the state tree, and the actions tree all have the _exact_ same structure.

taras: 🤔

cowboyd: It's like we have the `TypeTree` and the `InstanceTree`. The type tree is independent of value, but once you have value, the instance tree has `state`, `transitions`, `actions`, ..... and whatever other things still to come that proceed from it.

cowboyd: But they all have the same structure. I think it's that thing that we want to capture in a Functor to abstract away that structure.

taras: I'm not sure, but it feels like the right answer is tangled in how a microstate is constructed. An object that is an instance of `Microstate` is a wrapper for the object that has `state` and `transitions`. If we can clarify the distinction or role of these 2 types, then we might be able to make the Functor instance idiomatic.

PR #64, opened 2018-02-08
Structure Tree and Lenses

taras:
This PR changes what is returned inside of microstates when a transition is called. This is an internal refactoring, the external APIs remain unchanged. This refactoring was done in preparation for adding parameterized data types. 

Previously, a Microstates transition returned a Type and value. This required some acrobatics to allow type shifting. We were using a type lens that would create a new parent type when type shifting. This process was complicated and concerning because functions are not serializable. 

The changes introduced in this PR eliminate the need for all of the type acrobatics by allowing us to return a Structure tree instead of a Type. A structure tree represents each composed type as a node in the tree with Type and path. Working with structure trees allows us to perform type shifting by simply changing the value of Type property in the node's data. To make it easier to work with structure trees, we introduce tree lens. Tree lenses allow us to read and replace nodes on trees without manually managing and rebuilding trees. 

This PR also introduces a clarification to the responsibility of the Microstate class. The microstate class is now simply a DX API that wraps a Structure Tree. IT exposes transitions on the root of the microstate to make it easy to chain operations. 

The microstate function has now been renamed to `create`. The Microstate class now takes tree and value as arguments to the constructor.


taras:
@cowboyd ignore the build failure it's because of code coverage. I'll fix it in another PR

src/structure.js
```diff
@@ -0,0 +1,151 @@
...
+        value: localValue,
+        tree: prune(localTree)
+      };
+
+      let { 
+        value: nextLocalValue, 
+        tree: nextLocalTree 
+      } = invoke(transition);
+
+      if (nextLocalTree) {
```
cowboyd: I don't think we'll ever branch here, since we only ever have the next local tree now that `Type` has been removed from the invocation.

In fact, I think that the else will throw a `ReferenceError` if it is ever reached (which I don't think it will be) since `nextLocalType` is never declared.

PR #66, opened 2018-02-10
Run tests against dist package

taras:
0.5.0 failed because rollup wasn't configured properly. We could have prevented this problem if we were running our tests against the built dist package instead of source. This PR introduces a change to our test runner configuration to run tests against dist packages.

This this PR:
1. Properly configured rollup to declare all global and external dependencies
2. Build UMD and ES packages
3. Configure package.json to use UMD for main and browser and ES for module package entry points
4. Changed tests to use `microstates` module which forces Jest to go through `main` entry
5. Configured jest to rebuild the dist before running tests
6. To ensure that the `jest ---watch` works correctly, I'm rebuilding the dist files when any file changes 
7. Changed test coverage to only report `dist/microstates.umd.js` because thats the only file that's actually being exercised for most of the tests


rollup.config.js
```diff
@@ -1,22 +1,49 @@
-import babel from 'rollup-plugin-babel';
+import babel from "rollup-plugin-babel";
+import filesize from 'rollup-plugin-filesize';
+
+import pkg from "./package.json";
+const { keys } = Object;
+
+const globals = {
```
cowboyd: Having to explicitly enumerate all dependencies here in a rollup config seems fragile to me. Is this normally what you have to do?

taras: I don't know that much about "normal" use, but from what I understand, UMD is compatible with globals where ramda package is available on R namespace. In global mode, there is no way to declare where modules are, so you have to explicitly tell Rollup how to match imports to globally available functions.

taras: Btw, tests will now complain if something is missing from Globals because rollup happens when tests run.

taras:
@cowboyd so the tests are running against the package now but the coverage is completely messed up now.

PR #71, opened 2018-02-24
Adding sugar for parameterized data types

taras:
This PR adds sweet sweet sugar to parameterized data types. It allows the user to use `[Type]` and `{Type}` instead of `parameterized(Array, Type)` and `parameterized(Boolean, Type)`. 

It supports composing parameterized types, for example you can say `[{Number}]` or `[[Number]]` or `{[Number]}`. 

<details><summary> ~~There is one failing test, it looks like it might be related to how values are collapsed.~~ </summary>
screen shot 2018-02-24 at 6 59 08 am
</details>


.eslintrc.json
```diff
@@ -0,0 +1,23 @@
+{
```
cowboyd: Let's introduce this as a separate PR.

src/structure.js
```diff
@@ -15,22 +16,22 @@ export default function analyze(Type, value) {
   return flatMap(analyzeType(value), pure(Tree, new Node(Type, [])));
 }
 
-function analyzeType(value) {
+export function analyzeType(value) {
   return (node) => {
-    let InitialType = node.Type;
+    let SugarFreeType = isSugar(node.Type) ? desugar(node.Type) : node.Type;
```
cowboyd: Let's keep knowledge of sugaring inside the sugar module. Any reason why:

```js
let InitialType = desugar(node.Type);
```
won't work? if it's already desugard, then `desugar` can just passthrough.

src/structure.js
```diff
@@ -45,6 +46,7 @@ function analyzeType(value) {
           }
         }
         return $(new Type())
+          .map(value => isSugar(value) ? desugar(value) : value)
```
cowboyd: Again, we can hide the `isSugar` behind the `desugar` function. That let's us map cleanly:

```js
return $(new Type())
  .map(desugar)
```

src/structure.js
```diff
@@ -53,10 +55,32 @@ function analyzeType(value) {
   };
 }
 
-function isa(Child, Ancestor) {
+export function isa(Child, Ancestor) {
```
cowboyd: Is this used elsewhere?

src/structure.js
```diff
@@ -53,10 +55,32 @@ function analyzeType(value) {
...
   return Child === Ancestor || Child.prototype instanceof Ancestor;
 }
 
+export function collapseState(tree, value) {
+  return new $(tree)
+      .flatMap(node => {
```
cowboyd: So we're converting all simple values into leaf nodes? Effectively truncating their children, right?

src/structure.js
```diff
@@ -91,8 +115,8 @@ function graft(path, tree) {
   }
 }
 
-class Node {
-  constructor(Type, path, tree) {
+export class Node {
```
cowboyd: Does this need to be exported?

src/sugar.js
```diff
@@ -0,0 +1,35 @@
...
+const { keys } = Object;
+
+let ContainsTypes = Monoid.create(class ContainsTypes {
+  empty() { return true; }
+  append(a, b) {
+    // this is necessary because Funcadelic passes different arguments
+    // for arrays than objects. For arrays, value is just the value and 
+    // for Objects it's Monoidic with {value}. 
+    // I think it's a bug in Funcadelic.
+    let value = b && b.value || b;
```
cowboyd: As discussed in As discussed in https://github.com/cowboyd/funcadelic.js/issues/18, we should have a typeclass to convert values into lists if needed.

src/structure.js
```diff
@@ -53,10 +55,32 @@ function analyzeType(value) {
...
+          }
+          if (isa(node.Type, types.Object)) {
+            return analyzeType(value)(new Node(types.Object, node.path));
+          }
+        }
+        return analyzeType(value)(node);
+      })
+    .map(node => {
+      if (node.isSimple) {
+        return node.valueAt(value) || new node.Type(node.valueAt(value)).valueOf();
```
cowboyd: Can we not push this logic into `node.stateAt`? so that the map is just `map(node => node.stateAt(value)`?

src/structure.js
cowboyd: Above comment was to make sure I understand correctly mainly for my understanding

I think that a `flatMap` is perfect here and now I know what you were talking about. The incredible thing about this approach is that even though the entire tree is theoretically part of the intermediate representation, the parts that aren't part of the final representation will never be computed because of laziness.

.eslintrc.json
taras: moving

src/structure.js
taras: yep, good idea

src/structure.js
taras: done

src/structure.js
taras: > So we're converting all simple values into leaf nodes? Effectively truncating their children, right?

Yes.

src/structure.js
taras: No, removed export.

src/sugar.js
taras: done

PR #72, opened 2018-02-26
Observable Microstates and Observable.from interop

taras:
You can create an observable from any microstate using the Observable.from method. The resulting Observable will stream the next microstate for every transition. 

When you subscribe to the observable microstate, you'll imidiately receive a microstate throught the stream. This microstate will have transitions that you can call to cause the next microstate to come through the stream.

```js

let ms = create(Number, 42);
let observable = Observable.from(ms);

let last;

observable.subscribe(microstate => {
  last = microstate;
  console.log(microstate.state);
});
// => 42

last.increment().increment().increment();
// => 43
// => 44
// => 45
```

### DIFF
--- a/src/microstate.js
+++ b/src/microstate.js
@@ -1,6 +1,7 @@
 import { map } from 'funcadelic';
 import analyze, { collapseState } from './structure';
 import { keep, reveal } from './utils/secret';
+import SymbolObservable from 'symbol-observable'
 
 export default class Microstate {
   constructor(tree, value) {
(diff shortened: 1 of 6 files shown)


src/microstate.js
```diff
@@ -40,4 +41,26 @@ export default class Microstate {
     let { value } = reveal(this);
     return value;
   }
+
+  [SymbolObservable]() {
```
cowboyd: What do you think about putting this in its own module?

```js
import ObservableMicrostate from 'microstates/observable';

class Microstate extends ObservableMicrostate {
 ///
}
```

I don't know why, but I admire the tinyness of the Microstate class.

taras: Me too, but I think this is an import communication to people about microstates. If they look at the code, they can tell that hey can use it with an observable. I think it'll probably end up being the primary way of using microstates. We can always pull it out before 1.0.

What do you think?

cowboyd: Ok, I can see that. The main thing that I'm worried about is folks thinking it's the _only_ way to use microstates. But I'm ok with it.

PR #73, opened 2018-03-01
Introducing `static create` method for initialization

taras:
When modeling state machines, it's necessary to be able to initialize into a particular
state based on value. Previously, this was accomplished by returning a microstate from the constructor. This lead to exceeds maximum call stack when destination type extended from type shifting class. 

This PR introduces a new convention which allows the developer to describe how a state should be initialized for a specific type. `create` method must return a microstate that represents the state that you want the state to initialize into or nothing.  The create method takes one argument which is any `value`.

```js

class Session {
  content = null;

  static create(session) {
    if (session) {
      return create(AuthenticatedSession, session);
    } else {
      return create(AnonymousSession);
    }
  }
}

class AuthenticatedSession extends Session {
  isAuthenticated = true;
  content = Object;

  logout() {
    return create(AnonymousSession);
  }
}

class AnonymousSession extends Session {
  content = null;
  isAuthenticated = false;

  authenticate(user) {
    return create(AuthenticatedSession, { content: user });
  }
}
```

The create method follows the same semantics as the constructor. When you return a microstate, it'll use the type and value in the microstate for `this` node. If you return anything else, it'll ignore the result and create an empty instance of that type. 

One perk of this solution is that all types that extend from the root type will automatically get this create method.

### DIFF
--- a/src/structure.js
+++ b/src/structure.js
@@ -28,8 +28,12 @@ function analyzeType(value) {
     let instance = new InitialType(valueAt);
 
     if (instance instanceof Microstate) {
-      let { tree } = reveal(instance);
-      return graft(node.path, tree);
+      let { tree , value } = reveal(instance);
+      let shift = new ShiftNode(tree.data, value);
+      return graft(node.path, new Tree({
+        data: () => shift,
+        children: () => tree.children
+      }));
     } else {
       let Type = toType(InitialType);
       return new Tree({
@@ -183,5 +187,16 @@ class Node {
 
       return { tree: nextTree, value: nextValue };
     }, transitionsFor(Type));
+  }
+}
+
+class ShiftNode extends Node {
+  constructor({ Type, path }, value) {
+    super(Type, path);
+    assign(this, { value });
+  }
+
+  valueAt() {
+    return this.value;
   }
 }
--- a/src/structure.js
+++ b/src/structure.js
@@ -25,30 +25,31 @@ function analyzeType(value) {
   return (node) => {
     let InitialType = desugar(node.Type);
     let valueAt = node.valueAt(value);
+    let Type = toType(InitialType);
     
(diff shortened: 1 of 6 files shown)


Robdel12:
[woah](https://media1.tenor.com/images/b2190afe666915580adb0890260362d1/tenor.gif?itemid=4895431)

sivakumar-kailasam:
image

src/structure.js
```diff
@@ -25,21 +25,31 @@ function analyzeType(value) {
...
-        data: () => Type === node.Type ? node : append(node, { Type }),
-        children() {
-          let childTypes = childrenAt(Type, node.valueAt(value));
-          return map((ChildType, path) => pure(Tree, new Node(ChildType, append(node.path, path))), childTypes);
-        }
-      });
+      let { tree , value } = reveal(instance);
+
+      if (tree.data.Type === Type && value === valueAt) {
+        return tree;
```
cowboyd: Do we not need to graft this tree?

taras:
> A point on the naming of things which is, as we know, the hardest problem in programming. Should we call it initialize to disambiguate it from Microstates.create. Or is it similar enough that

I like the symmetry of `create` and we're describing how the state is created. What we're actually doing is creating the nodes, which happens at initialization stage and at all other times when object are created. So, initialization is a subset of all of the times when this function is necessary.

src/structure.js
taras: Yeah, you're right.

src/structure.js
```diff
@@ -33,7 +33,7 @@ function analyzeType(value) {
       let { tree , value } = reveal(instance);
 
       if (tree.data.Type === Type && value === valueAt) {
-        return tree;
+        return graft(node.path, tree);
```
cowboyd: We should definitely have a test for this codepath since it wasn't getting caught before. The symptom of not having the correct path is that `valueAt()` will be wrong.

PR #78, opened 2018-03-08
Getting valueOf from passed in object

### LINKED ISSUE #77: Simple way to pass microstate values downstream
Many microstates carry information forward as they progress through their state diagram. For example,  consider the state machine for a validation Rule. As it stands, we have to manually unbox the value of the current microstate to pass it down to the existing microstate.

```js

class Rule {
  input = String;
  description = String;

  get isIdle() { return true; }
  get isTriggered() { return false; }
  get isRunning() { return false; }
  get isFulfilled() { return false; }
  get isRejected() { return false; }

  setInput(input) {
    return create(Triggered, this.valueOf())
      .input.set(input);
  }
  reset() {
    return create(Rule, this.valueOf())
      .input.set('')
  }
}

class Triggered extends Rule {
  get isTriggered() { return true; }

  run() {
    return create(Running, this.valueOf());
  }
}

class Running {
  get isRunning() { return true; }

  fulfill() {
    return create(Fulfilled, this.valueOf());
  }

  reject(reason = "") {
    return create(Rejected, this.valueOf())
      .reason.set(reason);
  }
}

class Fulfilled extends Rule {
  get isFulfilled() { return true; }
}

class Rejected extends Rule {
  reason = String;
  get isRejected() { return true; }
}
```
By passing the `valueOf()` for the current microstate, the `input` property gets passed down the line to each subsequent state.

It would be nice if we automatically unboxed the value so that we could write (among other things) the `fulfill` method like so:

```js
  fulfill() {
    return create(Fulfilled, this);
  }
```

taras:
Closes #77

### DIFF
--- a/src/microstate.js
+++ b/src/microstate.js
@@ -1,7 +1,8 @@
-import { map } from 'funcadelic';
-import analyze, { collapseState } from './structure';
-import { keep, reveal } from './utils/secret';
-import SymbolObservable from 'symbol-observable'
+import { map } from "funcadelic";
+import analyze, { collapseState } from "./structure";
+import { keep, reveal } from "./utils/secret";
+import SymbolObservable from "symbol-observable";
+import isCallable from "./utils/is-callable";
 
 export default class Microstate {
   constructor(tree, value) {
@@ -18,6 +19,7 @@ export default class Microstate {
    * @param {*} value
    */
   static create(Type, value) {
+    value = value && isCallable(value.valueOf) ? value.valueOf() : value;
     let tree = analyze(Type, value);
     return new Microstate(tree, value);
   }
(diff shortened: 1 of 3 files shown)


src/microstate.js
```diff
@@ -18,6 +19,7 @@ export default class Microstate {
    * @param {*} value
    */
   static create(Type, value) {
+    value = value && isCallable(value.valueOf) ? value.valueOf() : value;
```
cowboyd: Is it possible for anything that is not `null` or `undefined` to not have a `valueOf()`

taras: @cowboyd not likely.

cowboyd: I guess my point is, is it merely sufficient to just check that `value != null`

taras: It's trippy that `(10).valueOf()` is 10 and `10.valueOf()` is an exception. 
screen shot 2018-03-09 at 8 15 34 am

PR #79, opened 2018-03-10
Introducing `typeclasses` directory and Collapse typeclass

taras:
Collapse typeclass is an operation that changes the structure of a hierarchical data structure. What it means for a type to be collapsed depends on the type itself. By defining a `Collapse` typeclass, we make it possible for a type to describe how it can be collapsed. The operation can be performed with the `collapse` function.

This PR does the following:
1. Introduce `typeclasses` directory (because we have a bunch of these now)
2. Defines Collapse typeclass
3. Moves `tree.collapsed` to an instance of Collapse typeclass for Tree
4. Refactors code and test to use `collapse` 
5. Moves Values typeclass to typeclasses directory


cowboyd:
I'm definitely cool with the typeclasses dir, but do wonder what advantage `collapse` provides over `fold`?

taras:
I guess with fold, you'd still have to have some collapse function that you need to pass to the fold. Instead of `collapse(tree)`, it would be `fold(collapse, tree)` because collapse takes a function as a first argument. No?

taras:
@cowboyd I'm not sure why coverage dropped. It fell by less than 0.01 of a percent. I guess some coveralls funkiness.

PR #80, opened 2018-03-13
Introducing `Microstate.from` to immutably transition arbitrary POJOs

taras:
`Microstate.create` allows to build an instance of a microstate from a Type and a value. Microstates will use this `Type` to build a data structure with all of its transitions and state. This works great when the data structure is known ahead of time and you use it enough to warrant declaring the structure of your data before calling `Microstate.create`. 

What if you have a POJO that you need to transition but you don't have its type? 

Before this PR, you were out of luck. This PR introducing `Microstate.from` which takes a value and builds a microstate that you can call transitions on. `Microstate.from` doesn't currently have any way to match custom types from your application but it can detect primitive values and provide transitions for that type.

For example, you can do

```js
import { from } from 'microstate';

from(42).increment().valueOf();
// => 43

from(true).toggle();
// => false

from('hello world').concat('!!!');
// => hello world!!!
```

You can also pass deeply nested POJOs.

```js
from({ a: { b: { c: 42 } } }).a.b.c.increment().valueOf();
// => { a: { b: { c: 42 } } }
```

You can do,

```js
from([[[false]]])[0][0][0].toggle().valueOf();
// => [[[true]]]
```

You can even combine them in whatever way you want,

```js
let incremented = from({
  products: [
    { title: "Legos" 
       blocks: [
         {shape: 'square', quantity: 20}
       ]
     },
  ]
}).products[0].blocks[0].quantity.increment();

incremented.valueOf();
// = > {
//  products: [
//    { title: "Legos" 
//       blocks: [
//         {shape: 'square', quantity: 21}
//       ]
//     },
//  ]
//}
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,7 +2,7 @@ import './typeclasses';
 import Microstate from './microstate';
 
 export default Microstate;
-export const { create } = Microstate;
+export const { create, from } = Microstate;
 
 export { default as Microstate } from './microstate';
 export { default as Tree } from './utils/tree';
--- a/src/microstate.js
+++ b/src/microstate.js
@@ -1,5 +1,5 @@
 import { map } from "funcadelic";
-import analyze, { collapseState } from "./structure";
+import analyze, { collapseState, analyzeFrom } from "./structure";
 import { keep, reveal } from "./utils/secret";
 import SymbolObservable from "symbol-observable";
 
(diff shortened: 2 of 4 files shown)


taras:
@cowboyd any objections to merging this branch?

cowboyd:
I still need to have a look at it.

taras:
Introduced by #113

PR #81, opened 2018-03-15
Make sure that state getter is only evaluted once for a microstate

taras:
This is a "low hanging fruit" type improvement that ensures the state getter value is changed for any microstate. This is to ensure that same microstate doesn't generate multiple instances of composed state at same path. This PR also add caching to getters.

### DIFF
--- a/src/microstate.js
+++ b/src/microstate.js
@@ -1,12 +1,20 @@
-import { map } from "funcadelic";
+import { map, append } from "funcadelic";
 import analyze, { collapseState } from "./structure";
 import { keep, reveal } from "./utils/secret";
 import SymbolObservable from "symbol-observable";
+import thunk from './thunk';
 
 export default class Microstate {
   constructor(tree, value) {
     keep(this, { tree, value });
-    return map(transition => transition, this);
+    
+    let state = thunk(() => collapseState(tree, value));
+
+    return append(map(transition => transition, this), {
+      get state() {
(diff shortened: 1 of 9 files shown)


src/microstate.js
```diff
@@ -1,12 +1,20 @@
...
 import { keep, reveal } from "./utils/secret";
 import SymbolObservable from "symbol-observable";
+import thunk from './thunk';
 
 export default class Microstate {
   constructor(tree, value) {
     keep(this, { tree, value });
-    return map(transition => transition, this);
+    
+    let state = thunk(() => collapseState(tree, value));
```
cowboyd: As of 0.3.2, `funcadelic` has a `stable` function, that we could potentially use. https://github.com/cowboyd/funcadelic.js/blob/master/src/stable.js#L1

cowboyd: But rather than making a stable thunk here, I think we should just make the semigroup instance of `Object` stable. In other words we map the appended properties, and if it's a getter, then we transparently convert it to a stable thunk here https://github.com/cowboyd/funcadelic.js/blob/master/src/semigroup/object.js#L9.

That would make the append be a simple:

```js
return append(map(transition => transition, this), {
  get state() {
    return collapseState(tree, value));
  }
});

taras: I have everything you described in a branch of funcadelic. I’ll make a PR

PR #82, opened 2018-03-16
Ensure value from `static create` is provided to composed children

taras:
This PR addresses an issue discovered when attempting to use `static create` method to initialize a parameterized array. The `static create` method was added to allow developers to control how a type is created. 

Expected: It should allow the developer to change the initialized type and value. The value of the microstate that is created in `static create` should be used for the `state` and `valueOf` the create microstate.

```js
class MatureStudent {
  canDrive = Boolean;
}

class Student {
  static create(attrs) {
    if (attrs > 21) {
       return create(MatureStudent, { ...attrs, canDrive: true });
    }
  }
}
```

In this example, I would expect `create(Student, { age: 40 }).state.canDrive === true`.

Actual: the value that was boxed in the microstate that was returned by `static create` was not being considered when create state of children. As a result, the composed states always their default state rather than the provided value.

## In this PR

This PR fixes this behavior by ensuring that value that's boxed in the returned microstate is 
1. applied to all of the grafted children
2. is available when building state for parameterized arrays and objects
3. is included in the valueOf on next transition

## Changes

1. `analyze` now returns `tree & value`.
2. when `static create` is used, the returned value is merged into valueOf using collapse value

## Notable discoveries.

1. `static create` is eager which prevents us from being able to make recursive `static create`
2. (I think) `valueOf` should become lazy and should only be materialized when `ms.valueOf()` is called
3. Location typeclass should probably include `valueOf`

# TODO
- [x] recursively shifting children
- [x] shifting values into nodes of parameterized data types 
- [x] collapse value to allow transitioning shifted parameterized data types

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6775,8 +6775,7 @@
     "stack-utils": {
       "version": "1.0.1",
       "resolved": "https://registry.npmjs.org/stack-utils/-/stack-utils-1.0.1.tgz",
-      "integrity": "sha1-1PM6tU6OOHeLDKXP07OvsS22hiA=",
-      "dev": true
+      "integrity": "sha1-1PM6tU6OOHeLDKXP07OvsS22hiA="
     },
     "staged-git-files": {
       "version": "0.0.4",
--- a/src/microstate.js
+++ b/src/microstate.js
@@ -28,7 +28,6 @@ export default class Microstate {
    */
   get state() {
     let { tree, value } = reveal(this);
-
     return collapseState(tree, value);
   }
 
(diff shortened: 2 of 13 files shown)


cowboyd:
Looking at this PR, I don't think that we can defer associating a (lazy) value with each node, since this effectively completes the circle of passing around {tree, value} pairs. There's literally nowhere that it isn't done, and if we're going to build proper abstractions for everything, then we just need to have access to the value at each point.

cowboyd:
I think we'll get some key wins if we do it this way:

1. `node.state` and `node.value` will always be a thing, so we'll get the laziness, _and_ stability we crave.
2. dramatically reduce complexity as we're just passing around the microstate structure everywhere.
3. because the tree is fully formed at all points, we can build abstractions around modification and structural sharing, especially with parameterized types.

taras:
@cowboyd should we merge this PR and make a new issue with your items from https://github.com/microstates/microstates.js/pull/82#issuecomment-374784253?

taras:
Changes in this branch are now part of `every-node-knows-its-value-and-state` branch.

PR #91, opened 2018-03-20
Fix Array transitions

cowboyd:
With the advent of parameterized types, the tree can extend _below_ the Array type, so it isn't enough just to work with the state or the value. We need to keep the tree in sync.

This implements a `splice` transition on the `ArrayType` which actually modifies the microstate tree to keep in sync with the value, and then `push`, `pop`, and `filter`, are implemented in terms of `splice`. Any nodes that can be shared are shared.

This is clearly not optimal, but we need to see what tree modification patterns look like for Object and others before abstracting.

### Notes

- In order to make module compilation work, I had to move `graft` and `prune` into the tree file.
- Because of cyclic dependencies, I had to get `create` and `Microstate` from the constructor. Again, this isn't ideal, but when we can find out what the right abstraction is, then the individual types should not need to know about Microstate at all.

### DIFF
--- a/src/structure.js
+++ b/src/structure.js
@@ -2,7 +2,7 @@ import $ from './utils/chain';
 import { type, map, append, pure } from 'funcadelic';
 import { flatMap } from './monad';
 import { view, set, lensTree, lensPath } from './lens';
-import Tree from './utils/tree';
+import Tree, { graft, prune } from './utils/tree';
 import transitionsFor from './utils/transitions-for';
 import { reveal } from './utils/secret';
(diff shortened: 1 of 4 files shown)


taras:
Looks good. 

Will this work with type shifting in map?

cowboyd:
> Will this work with type shifting in map?

No, not yet. That's another can of wormz to fry :)

taras:
I'm good to merge this. Should I do it?

cowboyd:
Not yet, I need to add some tests for `shift` and `unshift`

PR #100, opened 2018-04-13
Concentrate Lens logic in the lens module.

cowboyd:
We want to get to a point where we merge nodes and trees so that we can calculate complete things like states from directly within the node without any further context.

As a first step towards making that work, this removes lensing logic where it isn't needed, and moves ones that are still needed into the lensing module. That way, when we do need to make changes to the way the tree is changed, it can happen all in a single place.

### DIFF
--- a/src/lens.js
+++ b/src/lens.js
@@ -7,8 +7,11 @@ export { default as compose } from 'ramda/src/compose';
 
 import lens from 'ramda/src/lens';
 import view from 'ramda/src/view';
+import lset from 'ramda/src/set';
+import lensPath from 'ramda/src/lensPath';
 import Tree from './utils/tree';
-import { map, foldl } from 'funcadelic';
+import { append, map, foldl } from 'funcadelic';
+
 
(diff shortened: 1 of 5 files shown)


src/lens.js
```diff
@@ -7,8 +7,11 @@ export { default as compose } from 'ramda/src/compose';
 
 import lens from 'ramda/src/lens';
 import view from 'ramda/src/view';
+import lset from 'ramda/src/set';
```
taras: Why did you call this `lset` instead of just set?

cowboyd: Hahaha, this is the source of almost 2. hours of debugging.... It's because it conflicted with https://github.com/microstates/microstates.js/blob/050e2a2bfc9a28267d299f99cc5480363ba99a62/src/lens.js#L21 yet weirdly kinda worked in some situations.... so the error was not obvious.

Definitely a case where a static type system would have raised the issue immediately by recognizing that the two `set` functions had different signatures. HashTagLaughCry

Issue #123, opened 2018-05-28
Creating state fails in Node


taras:
`create(class Foo {}, {}).state` throws `TypeError: Class constructor Foo cannot be invoked without 'new'`. This can be seen in [this runkit](https://runkit.com/taras/problem-with-stabilizedclass). The problem seems to stem from the fact that [babel can't correctly extend native es6 classes](https://github.com/babel/babel/issues/4269#issuecomment-245413185).  

This can be easily reproduced by trying microstates in Node.js. 

<img width="1315" alt="screen shot 2018-05-28 at 8 07 27 am" src="https://user-images.githubusercontent.com/74687/40620678-85974396-624e-11e8-9593-870d1a6c676b.png">

taras:
It looks like the issue comes down to the fact that babel can't properly extend native classes. 

In Babel 7, they [introduced a wrapper](https://twitter.com/left_pad/status/940723982638157829) that can be applied to built-in classes. 

<img width="703" alt="screen shot 2018-05-28 at 10 48 41 am" src="https://user-images.githubusercontent.com/74687/40625015-b8dd5004-6264-11e8-9bf5-044ed684831b.png">
Source: https://medium.freecodecamp.org/were-nearing-the-7-0-babel-release-here-s-all-the-cool-stuff-we-ve-been-doing-8c1ade684039

TypeScript has a [proposal to introduce a helper to improve interop with ES5/ES6](https://github.com/Microsoft/TypeScript/issues/15397).

To work around this issue, the users have to either use native classes or babel classes. Which means that if we transpile classes in our library then app classes have to be transpiled. On the flip side, if native ES6 classes are used, then our library has to provide native ES6 classes. 

To do this, we can distribute `cjs` files for Node without transpiling classes.  We would only transpile the ES6 modules classes with the assumption that web apps would be transpiled with babel.

There is another alternative which is to remove class from the mechanism that stabilizes classes but I don't know how to achieve the same stable instances without the mechanism that we have.

cowboyd:
Is this still open?

cowboyd:
closing unless it's still a problem /cc @taras

PR #124, opened 2018-05-28
Don't transpile classes in CJS

taras:
This PR changes main to point to cjs built file which no longer transpiles es6 classes. This is done to prevent conflicts  between babel transpiler and native classes. ES6 classes remain unchanged.

This PR now includes UMD build that's linked to browser entry. This build has all dependencies necessary to use microstates. It's primarily for EmberCLI integration.

Related #123

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,9 +18,8 @@
     "dist",
     "README.md"
   ],
-  "main": "dist/microstates.umd.js",
-  "module": "dist/microstates.js",
-  "browser": "dist/microstates.umd.js",
+  "main": "dist/microstates.cjs.js",
+  "module": "dist/microstates.es.js",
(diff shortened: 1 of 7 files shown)


Robdel12:
Going to ping @wwilsman since he's had fun with modules in BigTest

cowboyd:
What are the environments that are effected?

cowboyd:
🤔  https://www.caniuse.com/#feat=es6-class

taras:
@cowboyd the assumption here is that the CJS will only be used by Node.js. All otherwise, will use the modules entry which transpiles for the browser. I'm this branch in a react app running on `create-react-app` and it's using the es modules build.

PR #132, opened 2018-06-06
WIP: better readme

taras:
Read the latest version of the [README.md](https://github.com/microstates/microstates.js/blob/tm/better-readme/README.md)


README.md
```diff
@@ -5,754 +5,360 @@
...
-// initial value is undefined so state will be default value of String which is an empty string
-create(String).state;
-// => ''
+When was the last time you had fun working with state in JavaScript applications? For many, the answer is probably never because state management in JavaScript is an endless game of choosing from compromises. You can choose to go fully immutable and end up writing endless reducers. You can go mutable and everything magically becomes an observable. Or you can `setState` and loose the benefit of serialization and time travel debugging. 
 
-// initial value is 'hello world' so state will be hello world.
-create(String, 'hello world').state;
-// => 'hello world'
-```
+Unlike the view layer, where most frameworks agree on some variation of React components, state management is alloot less certain. It's less certain because none of the available tools strike the balance that React components introduced to the view layer.
```
knownasilya: typo 'alloot'

taras: It's an inside joke for the Frontsiders ;)

README.md
```diff
@@ -5,754 +5,864 @@
...
 
-```bash
-npm install --save microstates
+Microstates is a functional runtime type system for declaratively describing immutable state transitions of deeply nested state objects designed to ease state management in component based applications.
 
-# yarn add microstates
-```
+By combining lazy execution, algebraic data types and structural sharing, we created
+a tool that provides a tiny API to describe complex data structures and provide a
+mechanism to change the value in immutable way.
```
knownasilya: 'in an immutable way'

README.md
```diff
@@ -5,754 +5,864 @@
...
-```js
-import { create } from 'microstates';
+With microstates added to your project, you get: 
 
-create(Number, 42).increment().state;
-//=> 43
+- 🍣 Composable hierarhical type system
+- 🍱 Extremely reusable state atoms 
+- 💎 Pure immutable state transitions without reducers
+- ⚡️ Lazy and syncronous by default
```
knownasilya: synchronous

README.md
```diff
@@ -5,754 +5,864 @@
...
+- 🔭 Optional integration with Observables
+- 🦋 Easiest way to express state machines
+- 🎯 Transpilation free type system
+- ⚛ Use in Node.js and web applications
 
-class MyCounter {
-  count = Number;
-}
+But, most imporantly, Microstates makes working with **state fun**.
```
knownasilya: 'importantly'
Only bold 'fun'

README.md
```diff
@@ -5,754 +5,864 @@
...
+But, most imporantly, Microstates makes working with **state fun**. 
 
-// creating microstate from composed types
-create(MyCounter).state;
-// => { count: 0 }
+## 🤗 Having fun with state 
 
-create(MyCounter).count.increment().state;
-// => { count: 1 }
+When was the last time you had fun working with state in JavaScript applications? For many, the answer is probably never because state management in JavaScript is an endless game of choosing from compromises. You can choose to go fully immutable and end up writing endless reducers. You can go mutable and everything magically becomes an observable. Or you can `setState` and loose the benefit of serialization and time travel debugging.
```
knownasilya: 'choosing between compromises'

README.md
```diff
@@ -5,754 +5,864 @@
...
 
-// you can restore a microstate from previous value
-create(MyModal, { isOpen: true, title: 'Hello World' }).state;
-// => { isOpen: true, title: 'Hello World' }
+React components have a tiny API, they are functional, simple and extremely reusable. The tiny API gives you high productivity for little necessary knowledge. Functional components are predictable and easy to reason about. They are conceptually simple, but simplicity hides architecture that makes them performant. Their simplicity, predictability and isolation makes them composable and reusable.
 
-create(MyModal, { isOpen: true, title: 'Hello World' }).isOpen.toggle().state;
-// => { isOpen: false, title: 'Hello World' }
+These factors combined is what makes React style components easy to work with and ultimately fun to write. Tiny API abstracting sophisticated architecture that delivers performance and is equaly useful on small and big projects is the outcome that we set out to achieve for state management with Microstates.
```
knownasilya: These combined factors are what make working with React style components easy

README.md
```diff
@@ -5,754 +5,864 @@
...
-create(MyType)
-  .counter.increment()
-  .modal.title.set('Hello World')
-  .modal.isOpen.set(true).state;
-// => {
-//  modal: { isOpen: true, title: 'Hello World' },
-//  counter: { count: 1 }
-// }
-```
+A Microstate is an JavaScript object that is created from a Microstate type. Microstate types are similar to JavaScript classes except instances created from Microstate types are immutable, which means that they can not be modified after they were created. To modify an immutable object, you must create a copy of the original object with the change that you wish to make.
```
knownasilya: A Microstate is **a** JavaScript object
after they **are** created

README.md
```diff
@@ -5,754 +5,864 @@
...
+let homer = create(Person, { firstName: 'Homer', lastName: 'Simpson' });
 
-ms.assign({ hello: 'world', hi: 'there' }).state;
-// => { hello: 'world', hi: 'there' }
+// original homer is unmodified
+let homerJ = homer.setFirstName('Homer J');
 ```
 
-## Composition
+Unlike a regular JavaScript class methods which imperatively mutate objects they have access to, Microstates types declaratively describe what should be different in the copy that's returned from the method.
```
knownasilya: Unlike regular JavaScript class methods, which imperatively mutate objects they have access to,

README.md
```diff
@@ -5,754 +5,864 @@
...
+
+### Getters
+
+Getters on JavaScript objects are evaluated everytime that a getter is read. This can very wasteful when properties of an object haven't changed. Frameworks like Ember and Vue have *computed properties* which memoize computations and automatically recompute when a property changes.
```
knownasilya: This can **be** very wasteful

README.md
```diff
@@ -5,754 +5,864 @@
...
+Unlike a regular JavaScript class methods which imperatively mutate objects they have access to, Microstates types declaratively describe what should be different in the copy that's returned from the method.
+
+### Getters
+
+Getters on JavaScript objects are evaluated everytime that a getter is read. This can very wasteful when properties of an object haven't changed. Frameworks like Ember and Vue have *computed properties* which memoize computations and automatically recompute when a property changes. 
 
-Microstates are designed to be composable. You can define custom microstate types and nest them as 
-necessary. Microstates will take care of ensuring that all transitioning composed microstates is
-done immutably.
+Tracking property changes is not possible in pure JavaScript, as a result there is no safe way to cache getters on JavaScript objects. This limitation of getters can lead to very suble performance problems when used in frameworks like React that use shallow equality to determine if components need to rerender.
```
knownasilya: suble -> subtle

README.md
```diff
@@ -5,754 +5,864 @@
...
-Microstate.create(Counter).state;
-// => { count: 0 }
+shop.filtered === shop.filtered
+//> false
 ```
 
-We use Class Properties syntax to define the composition of microstates. If you're using Babel, you need the
-[Class Properties transform](https://babeljs.io/docs/plugins/transform-class-properties/) for Class
-Properties to work.
+If you passed `shop.filtered` to a React component, this React component would re-render every time that the parent component's render function is called even though neither `products` nor `filter` is changed.
```
knownasilya: is changed -> have changed

README.md
```diff
@@ -5,754 +5,864 @@
...
+
+let restored = JSON.parse(JSON.stringify(shop.valueOf()));
+
+restored instanceof Shop
+//> false
 ```
 
-You can access transitions of composed states by using object property notation. Every time that you
-call a transition, you receive a new microstate and you start from the root of the original type.
+`shop.valueOf()` doesn't give you a primitive value, because `valueOf` class instance returns the shop object. It doesn't return a primivite value of the object, it returns the same object. `JSON.parse/JSON.stringify` are not enough because these functions don't know how to convert your POJOs into instances. If you look closely, you'll see that actually deserialization is done manually in the constructor. This gets very complicated with complex data structures and inpractical when you have complex relationships between data.
```
knownasilya: inpractical -> impractical

README.md
```diff
@@ -5,754 +5,864 @@
...
 
-```js
-create(Person).father.father.set({ name: 'Peter' }).state;
-// => { name: '', father: { name: 'Peter' }}
+shop.valueOf();
+// give you the value that can be used to restore the state
 ```
 
-`merge` will recursively merge the object into the current state.
+Microstates actually does deserialization by default. When you create a Microstate object, you provide the `create` function with the type and a POJO that represents the value. Microstates internally will create all of the instances using the type as a blueprint for the final state.
```
knownasilya: Microstates actually **do** deserialization by default.

README.md
```diff
@@ -5,754 +5,864 @@
...
-}
+## Why Microstates?
 
-create(Ajax).state.content;
-// => null;
+Out tools effect how we solve problems and collaborate. Two tools can serve the same purpose but foster a completely different kind of ecosystem. Take jQuery plugins and React components as an example. Both of these tools provided a way to add custom behaviour to a DOM element but they fostered very different communities. 
 
-create(Ajax).content;
-// undefined
+In React world today, we see many special purpose components that are easy to combine to create sophisticated user experiences. These components tends to have few options but provide just the right API for you to build what you need.
```
knownasilya: In **the** React world today
tends -> tend

README.md
```diff
@@ -5,754 +5,864 @@
...
-// => null;
+Out tools effect how we solve problems and collaborate. Two tools can serve the same purpose but foster a completely different kind of ecosystem. Take jQuery plugins and React components as an example. Both of these tools provided a way to add custom behaviour to a DOM element but they fostered very different communities. 
 
-create(Ajax).content;
-// undefined
+In React world today, we see many special purpose components that are easy to combine to create sophisticated user experiences. These components tends to have few options but provide just the right API for you to build what you need. 
 
-create(Ajax).state.isLoaded;
-// => false
+jQuery plugins on the other hand offer endsless list of options that are difficult get just right. In jQuery ecosystem, libraries tended to be monolithic because plugins from different libraries often did not work well together.
```
knownasilya: endsless list of options -> endless options
In **the** jQuery ecosystem
often **do** not work well together.

README.md
```diff
@@ -5,754 +5,864 @@
...
+In React world today, we see many special purpose components that are easy to combine to create sophisticated user experiences. These components tends to have few options but provide just the right API for you to build what you need. 
 
-create(Ajax).state.isLoaded;
-// => false
+jQuery plugins on the other hand offer endsless list of options that are difficult get just right. In jQuery ecosystem, libraries tended to be monolithic because plugins from different libraries often did not work well together. 
 
-create(Ajax).isLoaded.set(false);
-// Error: calling set of undefined
-```
+As a result, in React ecosystem we see components like [react-virtualized](https://github.com/bvaughn/react-virtualized) and [react-dnd](https://github.com/react-dnd/react-dnd) that are made by experienced React developers. These components save companies and developers millions of dollars in wasted development by eliminating the need for everyone to re-invent these components to build their user experiences.
```
knownasilya: in **the** React ecosystem

README.md
```diff
@@ -5,757 +5,886 @@
...
 
-## Intro
+<!-- tocstop -->
+</details>
 
-Microstates are a _typed_, _composable_ and _immutable_ state container.
+# Features
 
-Let's get some terminology out of the way so we're on the same page for the rest of the README.
+With microstates added to your project, you get:
```
sadatay: marking this for note below...

README.md
```diff
@@ -5,757 +5,886 @@
...
+* 🍱 Reusable state atoms
+* 💎 Pure immutable state transitions without writing reducers
+* ⚡️ Lazy and synchronous out of the box
+* 🦋 Most elegant way to express state machines
+* 🎯 Transpilation free type system
+* 🔭 Optional integration with Observables
+* ⚛ Use in Node.js and browser
 
-To add microstates to your project,
+But, most imporantly, Microstates makes working with state fun.
```
sadatay: we need to settle on a representation for Microstates that is consistent throughout the document.  above we use `microstates` but _mostly_ it seems to trend towards the proper nounified `Microstates`, which I'd prefer if given a choice between the two.  `microstate` does not seem to be a generic noun outside of the realms of physics and politics, so it seems worthy of the designation.  but i'm less hung up on which so much as that we're consistent throughout.

README.md
```diff
@@ -5,757 +5,886 @@
...
 
-```bash
-npm install --save microstates
+**When was the last time you had fun working with state?**
 
-# yarn add microstates
-```
+For many, the answer is probably never, because state management in JavaScript is an endless game of compromises. You can choose to go fully immutable and write endless reducers. You can go mutable and everything becomes an observable. Or you can `setState` and loose the benefit of serialization and time travel debugging.
```
sadatay: `and loose the benefit of` -> `and lose the benefits of`

README.md
```diff
@@ -5,757 +5,886 @@
...
 
-Let's see some code,
+Unlike the view layer, where most frameworks agree on some variation of React components, state management is a lot less certain. It's less certain because none of the available tools strike the balance that React components introduced to the view layer.
```
sadatay: `variation of React components` - would it be more accurate to say something like `variation of React's concept of components`?  I'm guessing most frameworks other than React aren't using actual React components.

sadatay: `...is a lot less certain. It's less certain...` - echo.  can change the opening of second sentence to something like `This is because` or join the sentences into one.  The former is probably preferable: joining would make this a bit long.

README.md
```diff
@@ -5,757 +5,886 @@
...
 
-```js
-import { create } from 'microstates';
+React components have a tiny API. They are functional, simple and extremely reusable. The tiny API gives you high productivity for little necessary knowledge. Functional components are predictable and easy to reason about. They are conceptually simple, but simplicity hides architecture that makes them performant. Their simplicity, predictability and isolation makes them composable and reusable.
```
sadatay: `Their simplicity, predictability and isolation makes them...` -> ` Their simplicity, predictability and isolation make them...`

knownasilya:
Coming along quite nicely!

README.md
```diff
@@ -5,757 +5,886 @@
 
 # Microstates
 
-* [Intro](#intro)
-* [microstate constructor](#microstate-constructor)
-* [Composition](#composition)
-* [Static values](#static-values)
-* [Computed Properties](#computed-properties)
+Composable State Primitives with a GraphQL inspired API.
```
cowboyd: I don’t think we should mention GraphQL here. This makes it feel way more heavyweight than it actually is. Composable State Primitives for JavaScript. I almost feel like we should lead with 

> Microstates is a functional runtime type system designed to ease state management in component based applications. It allows you to declaratively compose application state from atomic state machines.

How about:
> By combining lazy execution, algebraic data types and structural sharing, we created a tiny API to let you declaratively compose application state from immutable state machines.

Although, I feel though if we're going to mention state machines, then we should show a quick example of how they map together, because I don't think people (event David K. Piano) understand how they correlate.

README.md
```diff
@@ -5,757 +5,886 @@
...
-```bash
-npm install --save microstates
+**When was the last time you had fun working with state?**
 
-# yarn add microstates
-```
+For many, the answer is probably never, because state management in JavaScript is an endless game of compromises. You can choose to go fully immutable and write endless reducers. You can go mutable and everything becomes an observable. Or you can `setState` and lose the benefits of serialization and time travel debugging.
 
-Let's see some code,
+Unlike the view layer, where most frameworks agree on some variation of React's concept of components, state management is a lot less certain. This is because none of the available tools strike the balance that React components introduced to the view layer.
```
cowboyd: > Unlike the view layer, where most frameworks agree on some variation of React's concept of components, none of the current crop of state management tools strike the same balance that React components introduced to the API.

README.md
```diff
@@ -5,757 +5,886 @@
...
 
-class MyCounter {
-  count = Number;
-}
+It's not easy to find the right balance between simplicity and power, but considering the importance of state management in web applications, we believe it's a worthy challenge.
 
-// creating microstate from composed types
-create(MyCounter).state;
-// => { count: 0 }
+# Why Microstates?
```
cowboyd: When I see this. I'm like "Why Microstates? Is that what I'm reading right now?" I think this whole section should be in a blog post.

README.md
```diff
@@ -5,757 +5,886 @@
...
+blog.posts[0].state;
+//> Post<{ id: 1, title: 'Hello World' }>
+
+blog.posts[1].state;
+//> Post<{ id: 2, title: 'Most fascinating blog in the world' }>
 ```
 
-You can access transitions of composed states by using object property notation. Every time that you
-call a transition, you receive a new microstate and you start from the root of the original type.
+When you're working with parameterized types, the shape of the Microstate is determined by the value. In this case, `posts` is created with two items which will, in turn, create a Microstate with two items. Each item will be a Microstate of type `Post`. If you push another item onto the `posts` Microstate, it'll be treated as a `Post`.
```
cowboyd: It's better to just show, and let them infer. If there's more explaining that needs to happen afterwards, we can get into the subtleties aftewards.

In fact, I wouldn't talk about "parameterized types" just yet. That's not really a lever that we expose to control the API yet, so let's leave it out for now and just talk about "object microstates" and "array microstates"

README.md
```diff
@@ -5,757 +5,886 @@
...
+```js
+let blog2 = blog.posts.put("3", { id: 3, title: "It is only getter better" });
+
+blog2.posts["3"].state;
+//> Post<{ id: 3, title: 'It is only getter better' }>
 ```
 
-We currently do not support composition inside of arrays but we're looking into supporting it in the
-future.
+# State
```
cowboyd: Do we need to rewrite this to de-emphasize state?

README.md
```diff
@@ -5,757 +5,886 @@
...
+app.state.shop.products instanceof Array //> true
+app.state.shop.products[0] instanceof Product //> true
+app.state.shop.products[0].title === 'Sponge' // true
+app.state.shop.products[0].id === 1 //> true
 ```
 
-## Static values
+This illustrates that the object that's created from the state property is the same structure as the Microstate itself. According to its structure, each instance is populated by it's corresponding value from the value object.
+
+## All getters are cached
```
cowboyd: This is no longer true, right? Also, getters are now queries, so we may have to re-address this.

README.md
```diff
@@ -5,757 +5,886 @@
...
-// Error: calling set of undefined
+table.state.sortedRows === table.state.sortedRows
 ```
 
-## Computed Properties
+When you read this getter, it'll be evaluated and result will be cached. All future reads will return the same value.
+
+*Note*: It's ok if you've never done this before with regular JavaScript classes. JavaScript classes that were not instantiated with Microstates are mutable and do not cache their getters. A similar example without Microstates would cause the sort function to be invoked on each read.
+
+## Reuse of immutable state instances
```
cowboyd: What does this mean in the context of de-emphasizing state.

README.md
knownasilya: Can you elaborate on "de-emphasize state"?

README.md
taras: @sadatay I've been using `Microstates` when I refer to the library or the tool and `microstate` when I'm referring to an instance of the `Microstate` class

README.md
taras: Agree, but I need some advice on how to express it.

README.md
```diff
@@ -5,757 +5,886 @@
...
 
-ms.assign({ hello: 'world', hi: 'there' }).state;
-// => { hello: 'world', hi: 'there' }
-```
+When you want to change what the user sees, you could imperatively manipulate the DOM elements with jQuery, but React taught us to change DOM declaratively by changing the data that the DOM reflects. In the same way, when you want to change the state, you must change the value and allow the state to be reflected.
+
+# What is a Microstate?
+
+A Microstate is a JavaScript object that is created from a Microstate type and POJO value. The shape and methods of the Microstate are determined by its type and the value it contains. A type describes the structure and transitions that can be performed on the microstate. Using type and value, Microstates derives a single immutable state tree.
```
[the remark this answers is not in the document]
taras: We'll put a side by side todomvc example here

README.md
taras: @knownasilya Charles and spent the last 3 days re-thinking the role of getters in Microstates. There was previously a dichotomy between transitions and state. Getters were only on state and not on transitions because getters were used to automatically cache calculations on the state tree.

We had to roll back the caching of getters because extending JavaScript native classes with babel doesn't work. Between this and the fact that our structural sharing refactor changed the structure of microstates in a way that unlocked new capability, we're realizing that the state in it's previous conception might not be as useful.

State tree is immutable but when you pass it to components, you often need the transitions to make it useful. It's very rare that a component has readonly access to state. This got us thinking about how we could use getters better.

I now have a branch were we're intruding a feature that we talked about for a long time. It's a very small spike of what we're like to have ultimately, but it's enough to replace how we use getters.

This new feature is called querying. It allows to perform transitions but for the purpose of creating derivative state. Very similar to how use CPs in Ember the CP evaluates to another microstate that has transitions that you can invoke. The invoked transitions will be applied to the original objects that they reference in the original tree. 

This unlocks a lot of coolness that was not possible with previous state. With this in place, the state becomes less significant. The state property will still have the primitive value on it but its mostly to access value of the microstate.

README.md
```diff
@@ -5,757 +5,758 @@
...
+  towCapacity = Number;
 
-last.increment().increment().increment();
-// => 43
-// => 44
-// => 45
-```
+  tow(vehicle) {
+    if (vehicle.weight.state < this.towCapacity.state) {
+      return this.vehicle.set(vehicle);
```
wwilsman: I think this is meant to be `this.towing.set(vehicle)` because the `Vehicle` class defines it as `towing`.

README.md
```diff
@@ -5,757 +5,758 @@
...
-Microstates package provides base building blocks for your composing your own types. `Number`, `String`, `Boolean`,
-`Object` and `Array` come with predefined transitions.
+let prius = create(Car, { weight: 3000 });
 
-### `Boolean`
+let mustang = create(Car, { weight: 3500, towCapacity: 1000 });
+let f150 = create(Truck, { weight: 5000, towCapacity: 13000 });
 
-`Boolean` type presents a `true` or `false` value. `Boolean` has `toggle` and `set` transitions.
+let rob = create(Friend).vehicle.set(mustant);
```
wwilsman: Small typo: `mustang`

README.md
```diff
@@ -5,757 +5,758 @@
...
+
+from({ hello: [ 'world' ]}).hello[0].concat('!!!').valueOf();
+// { hello: [ 'world!!!' ]}
+```
+
+## map(fn: tree => tree, microstate: Microstate): Microstate
+
+The `map` function is used to perform operations on the tree that are used to build a new Microstate. It expects a function and a Microstate and returns a Microstate. This function accepts a mapping function which will receive the Microstate's tree. The mapping function is expected to return a tree. The tree that is returned from the mapping will be used to generate the microstate that's returned by the map operation.
+
+This is most frequently used go derive a microstate with middleware installed.
```
wwilsman: Another small typo: `frequently used to`

README.md
taras: Yes, you're right. Thank you. Good catch.

README.md
taras: Fixed.

README.md
```diff
@@ -609,27 +603,37 @@ from({ hello: [ 'world' ]}).hello[0].concat('!!!').valueOf();
...
 
-This is most frequently used go derive a microstate with middleware installed.
+```js
+let numbers = create([Number], [1, 2, 3, 4]);
+
+<ul>
+  {map(number => (
+    <li onClick={() => number.increment()}>{number.state}</li>
+  ), numbers)}
+</ul>
```
knownasilya: How would this pattern work in Ember, would `{{#each nubmer as |number|}}` be enough, or does `map` do something special?

taras: You should not need any mapping function as long as the microstate list object (array like) is compatible with the iterating mechanism in Glimmer. We might need to make a little tweak to microstates to ensure this compatibility. We will make it work with Ember without anything special. You'll be able to just use `each` without doing anything special.

README.md
```diff
@@ -5,757 +5,762 @@
...
-You can create an observable from any microstate using the `Observable.from` method.
-The resulting Observable will stream the next microstate for every transition.
-When you subscribe to the observable microstate, you'll immediately receive a microstate through the stream.
-This microstate will have transitions that you can call to cause the next microstate to come through the stream.
+Those familiar with functional programming might recognize this as a flatMap operation. It is not required for you to understand Monads to use Microstates transitions. If you're interested in learning about the primitives of functional programming, you may checkout [funcadelic.js](https://github.com/cowboyd/funcadelic.js). Microstates uses Funcadelic under the hood.
 
-```js
-import { create } from 'microstates';
-import { Observable } from 'rxjs';
+## Scope rules
```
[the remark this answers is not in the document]
cowboyd: Or even better "Transition scope"

README.md
knownasilya: I'm not sure @cowboyd. I'll give this a test.

README.md
```diff
@@ -5,757 +5,758 @@
...
 
-Defining custom types is done using [ES2016 classes syntax](http://exploringjs.com/es6/ch_classes.html). In JavaScript, classes are functions. So
-any function can work as a type, but we recommend using class syntax.
+# Types
 
-```js
-class Counter {
-  count = Number;
-}
+Microstates comes with 5 primitive types: `Boolean`, `Number`, `String`, `Object` and `Array`. In addition to primivite types, Microstates allows you to declare `class` types and two kinds of parameterized types: `[Type]` and `{Type}`.
```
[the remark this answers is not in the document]
taras: What kind of examples? Like this?`

```md
Microstates comes out of the box with 5 primitive types: `Boolean`, `Number`, `String`, `Object` and `Array`. With Just these basic types there's actually a lot you can do.

```js

let meaningOfLifeAndEverything = create(Number, 42);
meaningOfLifeAndEverything.state
//> 42

let greeting = create(String, 'Hello World');
greeting.state
//> Hello World

let isOpen = create(Boolean, true);
isOpen.state
//> true

let foo = create(Object, { foo: 'bar' });
foo.state
//> { foo: 'bar' }

let numbers = create(Array, [ 1, 2, 3, 4 ]);
numbers.state
//> [ 1, 2, 3, 4 ]
```
```

File: README.md [hunk 36]
```diff
@@ -5,757 +5,758 @@
...
+blog.posts["1"].state;
+//> Post<{ id: 2, title: 'Most fascinating blog in the world' }>
 ```
 
-## Computed Properties
-
-Composed states can have computed properties. Computed Properties make it possible define properties
-that derive their values from the state. Use getter syntax to define computed properties on composed
-states.
+Object microstates have `Object` transitions, such as `assign`, `put` and `delete`.
```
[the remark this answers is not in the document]
taras: I'll say: Object type microstates have `Object` transitions, such as `assign`, `put` and `delete`.

README.md
```diff
@@ -601,7 +603,7 @@ const lightMachine = Machine({
 });
 ```
 
-Here is equivalent state machine in Microstates,
+Microstates does not do any special state resolution. You explicitely declare what happens on a state transition. Here is what a similar state machine would look like in Microstates.
```
Robdel12: `explicitly` :)

taras: Thank you :)

README.md
```diff
@@ -5,757 +5,864 @@
...
-any function can work as a type, but we recommend using class syntax.
+Microstates brings the same kind of simplicity to the model layer. A Microstate is a functional model in a way that a component is a functional view. A component takes props and returns DOM; a Microstate takes value and returns state. DOM is a tree of element instances, state is a tree of model instances. A DOM tree is created by passing data to a root component. A state tree is created from a root type and value. A component is an abstraction that we use to manipulate the DOM tree. A microstate is an abstraction that we use to manipulate the state tree.
+
+When you want to change what the user sees, you could imperatively manipulate the DOM elements with jQuery, but React taught us to change DOM declaratively by changing the data that the DOM reflects. In the same way, when you want to change the state, you must change the value and allow the state to be reflected. -->
+
+# What is a Microstate?
+
+A Microstate is just an object that is created from a value and a type. The value is just data, and the type is what defines how you can transition that data from one form into the next.
```
cowboyd: I found myself wondering at about this point in the readme "what's the difference between a microstate and a pojo?" might be worth introducing that?

Unlike normal JavaScript objects, microstates are 100% immutable and cannot be changed. They can only derive new immutable microstates through one of their type's transitions.

README.md
```diff
@@ -5,757 +5,864 @@
...
+|                      +-age---+                    +-increment()->
+|                      |       | Microstate<Number> +-decrement()->
+|                      |       |                    +-set()->
+|                      |       |                    +-state: 39
+|                      |       +--------------------+
+|                      |
+|                      +-set()->
+|                      +-state: Person { name: 'Homer', age: 39 }
++----------------------+
 ```
```
cowboyd: I think it's worth calling out that the number microstate has transitions and that we'll see later how to invoke them, but worth introducing the idea that we don't have to write any.

README.md
```diff
@@ -5,757 +5,864 @@
...
-You can compose custom types into other custom types.
+## Composing types
 
-```js
-class Address {
-  street = String;
-  number = Number;
-  city = String;
-}
+Types can be composed with other types freely. This makes it possible to build complex data structures that accurately describe your domain. Since Microstates are atomic and all transitions return Microstates, Microstates can automatically handle transitions regardless of how your `class` types are composed.
```
cowboyd: This second sentence is a mouthful. Also, still has class type jargon. 

> Types can be composed with other types freely and Microstates will take care of handling the transitions for you. This makes it possible to build complex data structures that accurately describe your domain.

README.md
```diff
@@ -5,757 +5,864 @@
...
 
-class Person {
+Let's define another type that composes the person type.
```
cowboyd: ~composes~ uses.

README.md
```diff
@@ -5,757 +5,864 @@
...
+|                   |           +--------------------+
+|                   |
+|                   +-set()->
+|                   +-state: Car<{ designer: Person<{ name: 'Homer', age: 39 }> }, name: 'The Homer' }>
++-------------------+
 ```
 
-You can access transitions of composed states by using object property notation. Every time that you
-call a transition, you receive a new microstate and you start from the root of the original type.
+The property names are important when defining `class` types because they are used to reference composed microstates. You can use the object dot notation to access a composed microstate. Using the same example from above:
```
cowboyd: Let's de-jargon
> The type determines the properties of the microstate, and so you can use simple dot notation to access all of the microstates it contains.

README.md
```diff
@@ -5,757 +5,864 @@
...
+|                      |       |                    +-state: 39
+|                      |       +--------------------+
+|                      |
+|                      +-set()->
+|                      +-state: Person { name: 'Homer', age: 39 }
++----------------------+
 ```
 
-You can compose custom types into other custom types.
+## Composing types
```
cowboyd: Let's rename this to "Creating your own microstates"

README.md
```diff
@@ -5,757 +5,864 @@
...
-//      }
-//    }
+theHomerCar.designer.state;
+//> Person<{ name: 'Homer', age: 39 }>
+
+theHomerCar.designer.age.state;
+//> 39
+
+theHomerCar.name.state;
+//> The Homer
```
cowboyd: Might also be worth showing the whole homer car state

> // or we can access the top level state from the top level microstate!
> theHomerCar.state;
> //> { name: "The Homer", designer: {name: 'Homer', age: 39} }

README.md
```diff
@@ -5,757 +5,864 @@
...
+
+```js
+let blog2 = blog.posts.push({ id: 3, title: "It is only getter better" });
+
+blog2.posts[2].state;
+//> Post<{ id: 3, title: 'It is only getter better' }>
+```
 
-Composed states have two default transitions `set` and `merge`.
+## Object type Microstates
```
cowboyd: ~type~

README.md
```diff
@@ -5,757 +5,864 @@
...
-We currently do not support composition inside of arrays but we're looking into supporting it in the
-future.
+When you're working with parameterized types, the shape of the Microstate is determined by the value. In this case, `posts` is created with two items which will, in turn, create a Microstate with two items. Each item will be a Microstate of type `Post`. If you push another item onto the `posts` Microstate, it'll be treated as a `Post`.
+
+```js
+let blog2 = blog.posts.push({ id: 3, title: "It is only getter better" });
+
+blog2.posts[2].state;
+//> Post<{ id: 3, title: 'It is only getter better' }>
+```
```
cowboyd: It's worth calling out at the end of this section, something like

> Notice how we didn't have to do _any_ extra work to define the state transition of adding another post to the list? That's the power of composition!

Or something like that. The awesomeness of microstates is that you don't even notice how awesome it is. Folks might read that and take `ArrayType#push` for granted because "obvs, right?" No. Not obvs. That was true ultimate power you just witnessed.

README.md
```diff
@@ -5,757 +5,864 @@
...
+blog.posts[0].state;
+//> Post<{ id: 1, title: 'Hello World' }>
+
+blog.posts[1].state;
+//> Post<{ id: 2, title: 'Most fascinating blog in the world' }>
 ```
 
-We currently do not support composition inside of arrays but we're looking into supporting it in the
-future.
+When you're working with parameterized types, the shape of the Microstate is determined by the value. In this case, `posts` is created with two items which will, in turn, create a Microstate with two items. Each item will be a Microstate of type `Post`. If you push another item onto the `posts` Microstate, it'll be treated as a `Post`.
```
cowboyd: "The same is determined by the value." This is a statement as profound as it is true, but also a bit abstract for the first read through. Let's just say "Array microstates" for now and not use `parameterized`. That's an advanced topic and information that is not needed in the introductory material.

> When you're working with Array microstates, it ensures that all of its members are of the type you specified.

README.md
```diff
@@ -5,757 +5,864 @@
...
+blog2.posts["3"].state;
+//> Post<{ id: 3, title: 'It is only getter better' }>
 ```
 
-## Static values
+# Transitions
+
+Transitions are the operations that let you derive a new state from an existing state. All transitions return another Microstate.
+
+<img src="README/boolean-statechart.png" alt="Boolean Statechart" width="100" />
```
cowboyd: This image is really small. Also, it needs to be introduced:

> Consider this diagram describing a boolean state machine.

README.md
```diff
@@ -5,757 +5,864 @@
...
 
-You can define static values on custom types. Static values are added to state but do not get
-transitions.
+The simplest example of this is `toggle` transition on the `Boolean` type. The `toggle` transition takes no arguments and creates a new microstate with the state that is opposite of the current state. The `Boolean` type can be described with the following state chart.
+
```
cowboyd: > ~The simplest example of this is `toggle` transition on the `Boolean` type.~ The `toggle` transition ~takes no arguments and~ creates a new ~micro~state with ~the state that is opposite of the current state~ its boolean opposite. ~The `Boolean` type can be described with the following state chart.~

README.md
```diff
@@ -5,757 +5,864 @@
...
+Here is what this looks like with Microstates.
```
cowboyd: I'd include the actual definition that comes bundled, and _then_ usage.

README.md
taras: Yes, let's do it. I'll just add your snippet

README.md
taras: Done

README.md
taras: Done

README.md
taras: Done

README.md
taras: > Or something like that. The awesomeness of microstates is that you don't even notice how awesome it is. Folks might read that and take ArrayType#push for granted because "obvs, right?" No. Not obvs. That was true ultimate power you just witnessed.

This is blog post material

README.md
```diff
@@ -5,757 +5,875 @@
...
 ```
 
-## Changing structure
+## `initialize` transition
 
-When modeling state machines, it's often helpful to be able to initialize into a particular
-state based on value. This can be accomplished by adding a static `create` method. This method
-should return the next state that you want the node to initialize into.
+The `initialize` transition converts value into a Microstate when a Microstate is being created with the `create` function. The initialize transition is invoked for every microstate that declares one. They are called from top to bottom, meaning that a parent microstate is initalized before the children. This is imporant because the parent microstate can change what children are created.
+
```
cowboyd: This jumps into the what of the initialize transition, let's linger just a while on the Why.

> Just as every state machine must begin life in its "start state", so to must every microstate begin life in the right state. For those cases where the start state depends on some logic, there is the `initialize` transition. The `initialize` transition is just like any other transition, except that it will be invoked automatically within `create` on every Microstate that declares one.

README.md
```diff
@@ -5,757 +5,875 @@
...
+You can use this mechanism to change the value that is used to initialize children Microstates.
+
```
cowboyd: Capture the wonder!

> You can even use this mechanism to transition the microstate to one with a completely different type and value.

README.md
```diff
@@ -5,757 +5,875 @@
...
 ```
 
-## Static values
+# Transitions
+
+Transitions are the operations that let you derive a new state from an existing state. All transitions return another Microstate. You can use state charts to visualize microstates. The `Boolean` type can be described with the following statechart. 
 
-You can define static values on custom types. Static values are added to state but do not get
-transitions.
+<img src="README/boolean-statechart.png" alt="Boolean Statechart" width="250" />
```
cowboyd: This is still to small to read:

image

Can we put a caption on this?

README.md
```diff
@@ -5,757 +5,875 @@
...
 
-// transition to AnonymousSession state with logout
-create(MyApp, { session: { content: { name: 'Taras' } } }).logout().state;
-// => { session: { content: null, isAuthenticated: false }}
+result.vehicle.state.isTowing
+//> true
 ```
 
-## Parameterized Arrays & Objects
+Those familiar with functional programming might recognize this as a flatMap operation. It is not required for you to understand Monads to use Microstates transitions. If you're interested in learning about the primitives of functional programming, you may checkout [funcadelic.js](https://github.com/cowboyd/funcadelic.js). Microstates uses Funcadelic under the hood.
```
cowboyd: Nice shout out! Should we make this another "pro tip:"?

This is also an opportunity to express our core value of pragmatism and not letting the FP details "leak" into our implomentation.

> Pro tip: Microstates will never require you to understand Monads in order to use transitions, but if you're interested in learning about the primitives of functional programming that power Microstates, you may want to checkout [funcadelic.js](https://github.com/cowboyd/funcadelic.js).

README.md
```diff
@@ -5,757 +5,875 @@
...
-It's often useful to be able to indicate that an array consists of a certain type of items, 
-for example, an array of todo items. Microstates provides a special syntax for this.
+## Transition scope
 
-```js
-class Todo {
-  title = String
-  isCompleted = Boolean
-}
+For Microstates to be composable, they must work the same as a root Microstate or composed into another Microstate. For this reason, Microstate transitions only have access to their own transitions and the transitions of the Microstates that are composed into them. They do not have access to their context. This is similar to how components work. The parent component can render a child component and pass data to them. The children components do not have direct access to the parent component. Same in Microstates.
```
cowboyd: Let's strengthen this statement.

> Microstate are composable, and they work exactly the same no matter what other microstate they're a part of. For this reason.....

README.md
```diff
@@ -5,757 +5,875 @@
...
 
-Returns a new microstate with boolean value replaced. Value will be coerced with
-`Boolean(value).valueOf()`.
+## Immutable Object vs Immutable Data Structure
 
-```js
-Microstate.create(Boolean).set(true).state;
-// => true
-```
+When you create a state machine with xstate, you create an immutable object. When you invoke a transition on an `xstate` state machine, the value of the object is the ID of the next state. All of the concerns of immutable value change as a result of state change are left for you to handle manually.
```
cowboyd: This is a big problem, and should get a higher billing. We should drive in the point that we're marrying value and state because a state machine is worthless if you can't actually store values. I think an example of this would be really key. Maybe actually model the passage of time with a `tick` method?

README.md
```diff
@@ -5,757 +5,875 @@
...
-
-let last;
-
-observable.subscribe(microstate => {
-  last = microstate;
-  console.log(microstate.state);
-});
-// => 42
+```js
+const nextState = lightMachine.transition('green', 'TIMER').value;
```
cowboyd: In a way, microstates does have a transitionTo method. It's called "vanilla JavaScript property lookup"

```js
let next = lightMachine['timer']();
```

README.md
taras: I don't think so. I made the image 2x bigger.

README.md
taras: Can you think of an example for this? ( I can write around it )

README.md
taras: Yeah, how about this?

> Microstates does not have such a method. Instead, it relies on vanilla JavaScript property lookup. The method invocation is equivalent to calling `transitionTo` with name of the transition.

README.md
cowboyd: I think actually having the time tracking in there. Does it make the example too complicated?

README.md
cowboyd: I love it.

README.md
```diff
@@ -5,757 +5,877 @@
...
+|                   |           +--------------------+
+|                   |
+|                   +-set()->
+|                   +-state: Car<{ designer: Person<{ name: 'Homer', age: 39 }> }, name: 'The Homer' }>
++-------------------+
 ```
 
-You can access transitions of composed states by using object property notation. Every time that you
-call a transition, you receive a new microstate and you start from the root of the original type.
+You can use the object dot notation to access a composed microstate. Using the same example from above:
```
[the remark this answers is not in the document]
taras: I updated to use `sub microstates` instead of `composed microstate` throughout.

README.md
```diff
@@ -5,757 +5,877 @@
...
+
+Transitions are the operations that let you derive a new state from an existing state. All transitions return another Microstate. You can use state charts to visualize microstates. The `Boolean` type can be described with the following statechart. 
+
+<img src="README/boolean-statechart.png" alt="Boolean Statechart" width="500" />
+
+The `Boolean` type has a `toggle` transition takes no arguments and creates a new microstate with the state that is opposite of the current state. 
 
-You can define static values on custom types. Static values are added to state but do not get
-transitions.
+Here is what this looks like with Microstates.
```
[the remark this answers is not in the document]
taras: > I think we should show the state machine expression

What does this look like?

README.md
```diff
@@ -5,757 +5,875 @@
...
-Microstate.create(Boolean).state;
-// => false
+* [React.js](https://github.com/microstates/react)
+* [Ember.js](https://github.com/cowboyd/ember-microstates)
+* Create a PR if you created an integration that you'd like to add to this list.
+* Create an issue if you'd like help integrating Microstates with a framework
 
-Microstate.create(Boolean).toggle().state;
-// => true;
+# `microstates` npm package
```
Robdel12: I'm wondering if we should link to this part higher up in the README (like a link to "quick start"?)

README.md
```diff
@@ -5,757 +5,875 @@
...
+Then import the libraries using:
 
 ```js
-Microstate.create(Number).set(10).state;
-// => 10
+import Microstate, { create, from, map } from "microstates";
 ```
 
-#### sum(number: Number [, number: Number]) => microstate
+## create(Type, value): Microstate
```
Robdel12: While reading through this felt like it was missing an `# API` heading

knownasilya:
💯 Yay, great work guys!

PR #133, opened 2018-06-07
Added formatting to microstate types

taras:
This PR makes it possible display type as `Microstate<Type>` for microstates. It'll be helpful when people look at microstates in object inspectors like Redux Dev Tools.

screen shot 2018-06-06 at 9 41 26 pm

### DIFF
--- a/src/tree.js
+++ b/src/tree.js
@@ -38,7 +38,11 @@ const defaultMiddleware = (localMicrostate, transition, args) => {
 };
 
 export const transitionsClass = stable(function transitionsClass(Type) {
-  class Transitions extends Microstate {}
+  class Transitions extends Microstate {
+    static get name() {
+      return `Microstate<${Type.name}>`;
+    }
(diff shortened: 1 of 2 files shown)


cowboyd:
Immutable seems true, but perhaps overly vague. Why not just `MyApp`?

What about `Transitions` and `State`?

taras:
I thought about `Transitions` but they're really Microstates because they have `valueOf` and `state` properties. It's not just a transitions object. 

I could make it `State<MyApp>`. Is that better?

Robdel12:
> I could make it State<MyApp>. Is that better?

Yass! I like that a lot

taras:
screen shot 2018-06-07 at 9 32 41 am

This actually looks pretty good. What do you think?

Robdel12:
😍

taras:
I made the change, it now looks like the second image.

PR #137, opened 2018-06-18
[WIP] First pass at microstates lenses

keathleydavidj:
##Purpose
Right now microstates is using lenses from the Ramda utility library. Turns out implementing our own lenses is really straightforward, and this would allow us to cut Ramda out of the microstates bundle.

###Todos
[ ] Ensure type information is preserved


taras:
Should we turn this into a RIP PR?

PR #143, opened 2018-06-20
Release 0.8.0

taras:
Releasing 0.8.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.8.0",
   "description": "Graceful state transitions for data of any shape.",
   "keywords": [
-    "blood",
+    "statemachines",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,9 +1,9 @@
 {
   "name": "microstates",
-  "version": "0.7.3",
+  "version": "0.8.0",
   "description": "Graceful state transitions for data of any shape.",
   "keywords": [
-    "blood",
+    "statemachines",
```
cowboyd: let's add `lens` as a keyword while we're at it. And should it be `state-machine`?

taras: Sorry, I missed this. I'll add it in the README branch.

PR #145, opened 2018-06-21
Introducing Queries to Microstates

taras:
I'm going to attempt at writing this PR description second time since the first time GitHub ate my description. This PR is a bit of a bucket for all of the changes that Charles and I decided to make while attempting to write the README and create a TodoMVC implementation with latest Microstates.

## Queries

The biggest change introduced in this PR is the introduction of Queries to Microstates. Queries are computed properties that evaluate to a subset of a Microstate tree. It is useful when you to derive state from a Microstate.

We started down this path because we needed a way to create a filtered list of todos based on their status for the TodoMVC app. Previously. getters was used to derive state but they were only on the state tree. The getters were not applied to the microstate itself. This made it difficult to invoke transitions on the derived state. The queries APIs is much better than the previous getters because they are very consistent with transitions and can be used inside of transitions. 

Here is what TodoMVC filter query looks like,

```js
class TodoMVC {
  todos = [Todo]
  filter = String;
  get filtered() {
    switch(this.filter.state) {
      case 'show_completed': return this.todos.filter(todo => todo.completed.state).todos;
      case 'show_active': return this.todos.filter(todo => !todo.completed.state).todos;
      default: return this.todos;
    }
  }
}
```

Here are all of the things that you need to know about queries,

### Queries return Microstates

The goal of queries to extend a microstate with derived state. The derived state itself is a microstate. Inside of a query, you can perform transitions and return the result of a transition. This is what we're doing in the example above. The `todos.filter` transition returns a microstate with filtered results. You can also return primitive values. These values will be converted to microstates with `from`. This is to ensure consistent interface for interacting with properties on the microstate.

```js
class Person {
  firstName = String;
  lastName = String;
  get fullName() {
    return `${this.firstName.state} ${this.lastName.state}`;
  }
}

let homer = create(Person, { firstName: 'Homer', lastName: 'Simpson' });
homer.fullName.state
//> Homer Simpson
```

### Transition from Query result

An important feature of Microstate queries is the ability to invoke transitions on the result of the query. In the context of the TodoMVC example from above, this means that you can toggle the status of a todo item from the filtered list. The return from the transition will be the next microstate with the original microstate transitioned and the updated filtered list.

```js
let todomvc = create(TodoMVC, {
  filter: 'show_active',
  todos: [
    { title: 'Hello World', completed: false },
    { title: 'Greetings', completed: true },
    { title: 'Ola', completed: false }
  ]
});

let allDone = todomvc.filtered.map(todo => todo.completed.toggle());
allDone.filtered.length
//> 0
```

### Transitions inside of Queries

Being able to invoke transitions inside of queries makes it possible to consume the APIs that microstates provides to prepare the state for consumption outside of the microstate. Unfortunately, this currently has the limitation that you can only invoke transitions outside of microstates on microstates that exist in the microstate. IOW, you can't create microstates inside of the queries and invoke transitions on them outside of the queries. We'll likely make it possible in the future, once we figure out how this will work.

### Query results are memoized

In true Microstates fashion, Query results are memoized for stability and to reduce unnecessary work. This means that you can use queries from other queries and they will produce the same result in all queries. 

### Queries in transitions

You can consume the result of queries inside of transitions. For example, you could rewrite the TodoMVC example above to break up the filtered query into multiple queries.

```js
class TodoMVC {
  todos = [Todo]
  filter = String;

  get active() {
    return this.todos.filter(todo => !todo.completed.state).todos;
  }

  get completed() {
    return this.todos.filter(todo => todo.completed.state).todos;
  }

  get filtered() {
    switch(this.filter.state) {
      case 'show_completed': return this.completed;
      case 'show_active': return this.active;
      default: return this.todos;
    }
  }
}
```

### Array reduce transition

Array reduce is interesting because it's a first transition that we expect to reduce to a different type than an array. A common use case is to calculate the sum of all values or do other calculations. Here is an example of a reduce operation.

```js
class Calculation {
  numbers = Array;

  get sum() {
    return this
      .numbers.reduce((sum, item) => sum.increment(item.state), 0)
      .numbers;
  }

  get summary() {
    return `Sum of numbers is ${this.sum.state}.`
  }
}

create(Calculation, { numbers: [1, 2, 3, 4] }).sum.state
//> 10
```

### Array filter receives microstates

Previously, filter was unique in that filter function would receive state instead of a microstate. We've correct this mistake by making the filter function receive the microstate. This will allow filter to evaluate to some value based on a transition in of the passed in microstate.

```js
let list = create([Number], [ 10, 20, 30, 40 ]).filter(number => number.decrement(20).state > 0);
list.state
//> [30, 40]
```

This should cover all of the changes introduced in this PR.


src/tree.js
```diff
@@ -27,14 +27,23 @@ const { assign, defineProperties } = Object;
...
+    }
+  };
+
+  // transition is applied to the end of the path
+  // it returns the next tree which is returned
+  let nextTree = over(lens, applyTransition, root);
 
-  return microstate;
+  // map tree to make sure that it gets the root applied
+  return map(tree => tree, nextTree).microstate;
```
cowboyd: Tangential: I feel as though the requirement we have to maintain the root context within a tree is an albatross.

tests/examples/single-select.test.js
```diff
@@ -0,0 +1,127 @@
...
+
+    let isChildFocused = foldl((isChild, child) => {
+      return isChild ? true : focus.is(child);
+    }, false, values(tree.children));
+
+    if (transition.name === 'choose' && isChildFocused) {
+      return tree.microstate.map(selectable => {
+        let current = Tree.from(selectable);
+        if (current.is(focus)) {
+          return current.microstate.isSelected.set(true);
```
cowboyd: I feel as though this should be redundant. The transition happened at the site of the choice. We shouldn't have to deal with that case by default. What we're wanting to do is to reconcile the consequences higher up the tree with that transition. I don't fully understand this middleware yet, but I think it's a fault of our system if "pushing forward" a transition that has already been sanctioned is going to be a hard requirement.

taras: Agree, but our middleware is not complete. I can explain what's missing in person - requires diagrams.

taras:
> What is the major reason to want to perform transitions? Is it so that the filter, map, and find methods will live on the actual microstate instead of having to import them from microstates/query?

The biggest advantage of doing transitions in the query is the consistency that it introduces. Once you know how the transitions work, you can use them without having to keep 2 different map/filter/find functions in your mind. 

> It also feels strange that a todo item outside the query would not === the same item within the query. I can reconcile myself to the idea, but my instinct tells me that it is introducing layer of complexity that will be difficult to live with in the future.

I agree. This is much more complicated than I was hoping for it to be.

Now that I think about it, a simpler mental model for queries would be: "what's in the tree is what you get" This might be a lot simpler than the alternative in this branch. 

We could make `query` alias for Funcadelic chain and make all the operations on the chain operate on microstate without transitions. I would just implement type classes instances for map & filter. What do you think?

cowboyd:
This was the approach that I took with picostates just now, and it "just worked"

Getting these queries https://github.com/cowboyd/picostates.js/blob/master/tests/query.test.js#L18-L28

Was as simples as:

https://github.com/cowboyd/picostates.js/blob/master/src/array.js#L59-L70

taras:
Perfect!

PR #169, opened 2018-07-29
Export query functions so they can be used

taras:
Currently, queries are not exposed which makes it impossible to use them in apps. 

Changes in this PR:

* exposes `reduce` and `filter` functions from main package to allow `import { reduce, filter } from 'microstates'`
* removed `sum` function and added some smoke tests to make sure that all exports have asserts
* ~added Mappable typeclass and Mappable instance to Array type~

### DIFF
--- a/src/query.js
+++ b/src/query.js
@@ -11,12 +11,4 @@ export const Reducible = type(class Reducible {
   }
 });
 
-export const Mappable = type(class Mappable {
-  map(mappable, fn) {
-    let { map } = this(mappable);
-    return map(mappable, fn);
-  }
-});
-
-export const { map } = Mappable.prototype;
 export const { reduce, filter } = Reducible.prototype;
--- a/src/query.js
+++ b/src/query.js
@@ -11,4 +11,12 @@ export const Reducible = type(class Reducible {
   }
 });
 
+export const Mappable = type(class Mappable {
+  map(mappable, fn) {
+    let { map } = this(mappable);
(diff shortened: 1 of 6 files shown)


src/query.js
```diff
@@ -6,13 +6,17 @@ export const Reducible = type(class Reducible {
...
-    return reduce(reducible, (total, member) => total + getter(member), 0);
-  }
-
   filter(reducible, predicate) {
     return reduce(reducible, (filtered, member) => predicate(member) ? filtered.concat(member) : filtered, []);
   }
 });
 
-export const { reduce, sum, filter } = Reducible.prototype;
+export const Mappable = type(class Mappable {
```
cowboyd: How is this different than Functor?

tests/package.test.js
```diff
@@ -0,0 +1,13 @@
+import expect from 'expect';
+
+const pkg = require('../dist/microstates.es.js');
```
cowboyd: Could we not just make these assertions on the index?

src/query.js
taras: Functor should return the same type as the original functor. Mappable maps to array. In context of React, Functor instance can not be mapped to components.

tests/package.test.js
taras: We can, but I really want a smoke test to ensure that the built assets have the export.

taras: I changed this as well.

PR #170, opened 2018-07-29
Add Iterator to ArrayType

taras:
This PR introduces the ability to treat ArrayTypes as iterators. This allows to use the spread operator and provide the ability to map Microstate Arrays to React Components.

 Question: should I add a Functor Instance to Array type in this PR?


cowboyd:
Why not use a GeneratorFunction here? Then you get next and done API for free

PR #172, opened 2018-07-29
Added map query to Reducible

taras:
This PR makes it possible to use map operation in queries. It will create an array object with a function applied to each item. This should make it possible for us to map an array of microstates to React Components.

```jsx
import React from 'react';
import { map, create } from 'microstates';

let numbers = create([Number], [1, 2, 3]);

React.render(root, (
  <ul>
    {map(numbers, number => (
      <li>
       <button onClick={() => number.increment() })>Increment {number.state}</button>
     </li>
    ))}
  </ul>
));
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,4 @@
 export { create } from './src/microstates';
 export { ArrayType, ObjectType } from './src/types';
 export { default as from } from './src/literal';
-export { reduce, filter } from './src/query';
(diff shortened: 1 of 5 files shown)


taras:
@cowboyd we need it in a situation where `reduce` query operates on a value that can be either an ArrayType or a regular Array.

PR #189, opened 2018-08-03
Make DSL expansion 100% deterministic.

### LINKED ISSUE #186: Order of imports matters with syntactic sugar
Syntactic sugar is broken for Arrays and Objects due to an order-of-operations problem with the definition of different sugars for object and Array being in different places.

expected

```
> create([])
Microstate<Array<Any>> { state: [] }
```

actual:

```
> create([])
Microstate<Constant> { state: [] }
```

This happens because the list of sugar functions is mutable and depending on how modules are imported, different sugar transforms get put at different orders. :(

It should be made deterministic so that order does not matter.

cowboyd:
The "sugar" package allowed for any module to register its own syntactic transformations. This prevented circular dependiencies and added a theoretical advantage in that the microstates syntax could be extended. While this _might_ be something we want in the future, it also introduced some non-determinism since the order in which syntactic transformations were registered resulted in different outcomes.

This completely rewrites the syntactic transformation system to be 100% deterministic and located in a single place. (dsl.js)

It's very clear what the precedence of rules are, and what _all_ the rules are just by looking in a single place.

In order to test that syntactic sugar is working, many of the testcases have been converted to use it rather than instantiate the types directly.

This fixes #186


src/dsl.js
```diff
@@ -0,0 +1,90 @@
...
+    case Object:
+      return { Type: ObjectType, value: {} };
+    case Array:
+      return { Type: ArrayType, value: [] };
+    case Boolean:
+      return { Type: BooleanType, value: false };
+    case Number:
+      return { Type: NumberType, value: 0 };
+    }
+    // TIL switch doesn't work on String ¯\_(ツ)_/¯
```
knownasilya: 😮

PR #193, opened 2018-08-03
Implicitly bind methods for microstates inside a store.

### LINKED ISSUE #176: Transitions can not be invoked without their context
It's useful to be able to invoke a transition without their microstates. For example, when you want to [destructure a transition](https://runkit.com/taras/5b60b1d8f0b9bb001263ed91) from it's context or [bind transitions to event handlers](https://codesandbox.io/s/qx8pzov176).

cowboyd:
When making a user interface, you often need to pass the call back as an atomic function. For example, in React you would say:

```jsx
<MyComponent onClick={() => counter.increment() }></MyComponent>
```

With this change, the function property of every microstate is implicitly bound to that microstate so that the following code can be written as:

```jsx
<MyComponent onClick={counter.increment}></MyComponent>
```

resolves #176 

### Open Questions:

- We should probably clean up and unify all of the muckery that happens on prototypes just to cut down the complexity and code-size.
- By doing this, we're making stack traces a little more weird.
- Should we also make implicitly bound transitions on the raw microstates.

### DIFF
--- a/src/bind-methods.js
+++ b/src/bind-methods.js
@@ -1,0 +1,29 @@
+import { filter, map } from 'funcadelic'
+
+export default function bindMethods(Type) {
+  var descriptors = Object.getOwnPropertyDescriptors(Type.prototype);
+
+  let methodDescriptors = filter(({ key: name, value: desc }) => {
+    return name !== 'constructor' && typeof desc.value === 'function';
+  }, descriptors);
+
+  let boundMethodDescriptors = map((desc, name) => {
+    return {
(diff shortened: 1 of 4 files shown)


taras:
> We should probably clean up and unify all of the muckery that happens on prototypes just to cut down the complexity and code-size.
> By doing this, we're making stack traces a little more weird.

It's probably worth it right?

> Should we also make implicitly bound transitions on the raw microstates.

I don't know yet. Let's see. We can always add it if necessary.

PR #212, opened 2018-09-02
Support non-transitions

TeknoFiend:
Perhaps this violates the ideal of this project, but I think in some cases it is useful to have methods that are not transitions (or at least have the ability to return a value other than the `microstate` state). For simplicity the test-case in this PR is something that could be accomplished with a `getter` (which do not appear to be transformed into transitions).  Our actual use-case is a bit more complex but I could describe it if it helps explain why this would be useful.

The `nonTransition` marker would be more conveniently implemented as a decorator but I opted to not use them in the test-case because it's not a settled feature and the change in proposed syntax makes things complicated there.

The `nonTransition` name is just a placeholder.  If this idea is palatable we can choose an ideal name.

### DIFF
--- a/src/microstates.js
+++ b/src/microstates.js
@@ -68,7 +68,7 @@ const toMicrostateType = stable(function toMicrostateType(Type) {
   let descriptors = Object.getOwnPropertyDescriptors(Type.prototype);
   let methods = Object.keys(descriptors).reduce((methods, name) => {
     let desc = descriptors[name];
-    if (name !== 'constructor' && name !== 'set' && typeof name === 'string' && typeof desc.value === 'function') {
+    if (name !== 'constructor' && name !== 'set' && typeof name === 'string' && typeof desc.value === 'function' && !desc.value.nonTransition) {
       return methods.concat(name);
     } else {
       return methods;
--- a/tests/microstates.test.js
+++ b/tests/microstates.test.js
@@ -159,4 +159,23 @@ describe("Microstates", () => {
       });
     });
   });
+
+  describe('Microstates with nonTransition methods', function() {
+    class Person {
+      firstName = [String]
+      lastName = [String]
+      fullName() {
+        return this.firstName.state + " " + this.lastName.state;
+      }
+    }
+    Person.prototype.fullName.nonTransition = true;
+
+    let person;
+    beforeEach(function() {
+      person = create(Person, { firstName: 'Homer', lastName: 'Simpson' });
+    });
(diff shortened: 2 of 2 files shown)


mharris717:
Not exactly the same, but you can make a getter that return a function to achieve a similar thing today.

```
class Person {
  firstName = [String]
  lastName = [String]
  get nameWithTitle() {
    return (title) => {
      `${title} ${this.firstName.state} ${this.lastName.state}`
    }
  }
}
```

cowboyd:
TBH, we've toyed with the idea of adding a `@method` decorator to accomplish just this, but we've thus far shied away from it instead either opting to go with the option above, or most often, have a `helper` function to do the thing that you need.

What's the particular use-case?

TeknoFiend:
The original use-case was somewhat ill-conceived for different reasons but is seemed a natural enough concept that I submitted the PR anyway.  I looked around a bit and came up with a toy example (loosely based on real code we have for shipping):

```
class Square {
  length = Number
  width = Number

  @method
  canItemFit( length, width ) {
    return (length <= this.length && width <= this.width) ||
           (width <= this.length && length <= this.width);
  }
 }
```

You could definitely solve these cases with helpers, but IMHO that's not always the ideal solution. For relatively small bits of logic, that only apply to the object type, I personally want the logic located with the object because of the simplicity.

As I said this is not a hot topic for me at the moment.  And I agree that using `get`ers works for many, if not most, cases.

Thanks for the consideration.

cowboyd:
Another potential approach:

```js
function Pack(length, width) {
  return class Pack {
    square = Square    
    get willFit() {
      let sLength = this.square.length.state;
      let sWidth = this.square.length.state;
      return (length <= sLength && width <= sWidth) ||
           (width <= sWidth && length <= sWidth);
      }
  }
}

let pack = create(Pack(5,2), {square: {width: 5, length: 10})
pack.willFit //=> true
```

This has the advantage of being a stable property that can be used to render as a model.

This is a bit cumbersome at the moment, but when we implement microstate variables, you'd be able to test multiple `Packs` against a single square, or multiple squares against as single `Pack` which might actually make handling the combinatorics more elegant.

taras:
@TeknoFiend I'm going to close this PR because I don't think it has a future of being merged. There are a lot of good suggestions here for alternative implementations. Please, feel free to open an issue to discuss this further or reach out to me directly to discuss it. 

Thank you for you contribution, sorry it didn't make it in 🙃

TeknoFiend:
👍

PR #213, opened 2018-09-09
NOT A REAL PR - ArrayWrapper basically works outside of Identity

mharris717:


cowboyd:
I think I see what you're trying to accomplish here? Is the goal here to have certain microstate classes behave like arrays?

If so, I think you're going to be pleased to know that we're currently working to have the query functions work generically on any JavaScript [iterable object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#The_iterable_protocol) So all you would need to do is implement that, and be good to go.

cowboyd:
> Does working on any Iterable replace using the Reducible type?

Yes, it eventually will. All you'll have to do is implement the vanilla JavaScript iteration protocol, and you can use `map`, `reduce`, `filter`, and friends for free.

cowboyd:
With the advent of Microstates 0.11.0, Array microstates now support `Symbol.iterator`, and so can be used anywhere you would use a normal array.

PR #225, opened 2018-09-24
added map & filter to ObjectType

### LINKED ISSUE #224: Map for Object Microstates
Do you have any plans to add a `map` transition to Object microstates? I have a collection of values that i want to store in an object keyed by ids so I get O(1) value lookup. However, microstates doesn't seem to currently support functor operations on Objects. I can work around it with a forEach loop, and it's not like the difference of 1 line of code vs 5 is a huge deal. However, microstate code is highly functional in style already, so it would be nice if this was supported too.

<details>
<summary>
Example Code (https://codesandbox.io/s/vv709w87v5)
</summary>

```js
class Todo {
  id = create(String);
  completed = create(Boolean, false);
}

class MyModel {
  todosById = create({ Todo });

  completeAll() {
    let newValue = this;

    // I wish there was a functional way to do this
    // e.g. this.todosById.map(todo => todo.completed.set(true));

    Object.keys(this.todosById.state).forEach(id => {
      newValue = newValue.todosById[id].completed.set(true);
    });

    return newValue;
  }
}
```
</details>

P.S. I tried making a PR to add it myself, but due to my inexperience with functional lenses I couldn't quite make it work. I tried something like the following, but then I end up with each item in the dictionary belonging to it's own world due to the way lenses return the context object. 

```js
// src/types/object.js
  map(fn) {
    return map((value, key) => fn(this[key], key), this.state);
  }
```

I wanted to map over `this` rather than `this.state`, but when I did that I ended up mapping over extra keys like 'state' that aren't actually part of my user land object.

rkeeler:
Adds support for map, filter, and reduce transitions on Object microstates as discussed in https://github.com/microstates/microstates.js/issues/224.

Closes #224

### DIFF
--- a/src/types/object.js
+++ b/src/types/object.js
@@ -1,6 +1,6 @@
 import { append, filter, map } from 'funcadelic';
 import parameterized from '../parameterized'
-import { mount, valueOf } from '../meta';
+import { mount, valueOf, sourceOf } from '../meta';
 import { create } from '../microstates';
 
 export default parameterized(T => class ObjectType {
@@ -34,5 +34,13 @@ export default parameterized(T => class ObjectType {
 
   delete(name) {
     return filter(({ key }) => key !== name, valueOf(this));
+  }
+
+  map(fn) {
+    return map((v, key) => fn(sourceOf(this[key]), key), valueOf(this));
+  }
+
+  filter(fn) {
+    return filter(({ key }) => fn(sourceOf(this[key]), key), valueOf(this));
   }
 });
--- a/src/types/object.js
+++ b/src/types/object.js
@@ -37,7 +37,7 @@ export default parameterized(T => class ObjectType {
   }
 
   map(fn) {
-    return map((v, key) => fn(sourceOf(this[key]), key), valueOf(this));
+    return map(value => valueOf(fn(create(T, value))), valueOf(this));
   }
 
   filter(fn) {
--- a/src/types/object.js
+++ b/src/types/object.js
@@ -1,6 +1,6 @@
 import { append, filter, map } from 'funcadelic';
 import parameterized from '../parameterized'
-import { mount, valueOf, sourceOf } from '../meta';
+import { mount, valueOf } from '../meta';
 import { create } from '../microstates';
 
 export default parameterized(T => class ObjectType {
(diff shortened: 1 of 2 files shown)


cowboyd:
This is a great addition!

One note, I don't think we want to actually have a "reduce" transition on the microstate itself since that would represent a transition from an array microstate to a scalar microstate, since transitions should always return a new microstate.

cowboyd:
@rkeeler Just wanted to check in and see if you have any questions about migrating this PR to the microstates 0.11.0 api. 

I'd love to see this get included, so please let me know if there are any blockers.

taras:
@rkeeler can you please update the PR to remove `reduce` and base it? Otherwise it looks great.

src/types/object.js
```diff
@@ -35,4 +35,12 @@ export default parameterized(T => class ObjectType {
   delete(name) {
     return filter(({ key }) => key !== name, valueOf(this));
   }
+
+  map(fn) {
+    return map((v, key) => fn(sourceOf(this[key]), key), valueOf(this));
```
cowboyd: It might be conceptually (and computationally) simpler to just create the mapping context using `create` directly instead of `sourceOf` like we do in the `ArrayType`.

if we access the substate via the parent state we're reifying the substate (via create), then mounting it, then getting the source of the mounted substate, and using it as the mapping context.

If on the other hand, we just create directly, we skip everything except the create portion.

Another key advantage is that the `sourceOf` API remains private to the metadata. Not only do we have to not commit to maintaining this API, but it means less concepts for authors who want to implement their own parameterized types.

```suggestion
    return map(value => valueOf(fn(create(T, value))), valueOf(this));
```

Of course, if the property already _is_ reified, then it's going to be fast. 

🤷‍♂️ Hard to say which is going to be more often the case without data, so I'd opt for simplicity here.

rkeeler: Sounds good. By leveraging the public API, it is certainly a lot easier to understand what's going on there. I applied your suggested change.

src/types/object.js
```diff
@@ -35,4 +35,12 @@ export default parameterized(T => class ObjectType {
   delete(name) {
     return filter(({ key }) => key !== name, valueOf(this));
   }
+
+  map(fn) {
+    return map(value => valueOf(fn(create(T, value))), valueOf(this));
+  }
+
+  filter(fn) {
+    return filter(({ key }) => fn(sourceOf(this[key]), key), valueOf(this));
```
cowboyd: By the same token, I don't think we need `sourceOf` here, and instead can opt to just create a one-off filter context.

rkeeler: Sounds good. I fixed that and took out the import for `sourceOf`

taras:
@rkeeler thank you for your contribution 👍

PR #229, opened 2018-10-09
Rework opening line.

cowboyd:
There's a lot of buzzword in the opening line of the readme that might
scare folks away.

Now that I've had a lot of conversations with folks about what excites them personally about microstates, I believe the most compelling arguments are actually pretty simple.

1. Benefits of immutability, without the overhead (Think mutably, act immutably.)
2. performance via stable objects.
3. dead simple re-use via composition.

The opener talks a lot about the mechanics, but I worry that that can obfuscate those key benefits.

This addresses those benefits head-on, letting the rest of the README take care of the how.

Alternatives
===

- Microstates let you reap the benefits of strict immutability, yet keep the simplicity of the standard JavaScript object model.
- Microstates: pure functions over immutable data doesn't have to be hard.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,9 +6,8 @@
 
 # Microstates
 
-Microstates is a functional runtime type system designed to ease state management in component based applications. It allows you to declaratively compose application state from atomic state machines.
-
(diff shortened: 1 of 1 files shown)


keathleydavidj:
This is a fantastic change.  Simple, straightforward language that any JavaScript engineer will understand.
👍

PR #230, opened 2018-10-09
Update documentation for missing state.

cowboyd:
With the advent of the femtostates architecture, the `state` property is no longer on composite microstates (those microstates which are made out of combining other microstates).

Now, the only `.state` property is on primitive microstates (Number, String, Any, Constant)

This updates the docs to use the `valueOf` which is the new preferred method for extracting the aggregate value out of a microstate.

Also, Array microstates no longer allow direct, keyed access. Instead, the docs now use iteration for the demos.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ extending Person to have a `set()` method:
 |                      |       +--------------------+
 |                      |
 |                      +-set()->
-|                      +-state: { name: 'Homer', age: 39 }
+|                      |
 +----------------------+
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -250,10 +250,13 @@ theHomerCar.name.state;
 //> The Homer
 ```
 
-The state from sub microstates is also available on the parent object's state.
+To retrieve the underlying value represented by a microstate, you can
```
taras: Maybe change the order of this?

> You can use `valueOf()` function available from the microstates module to retrieve the underlying value represented by a microstate.

PR #234, opened 2018-10-11
Adjust bundle size in README.

cowboyd:
With the movement to a fully lazy architecture, we've reduced the web bundle size by about 20% to 3.78KB in the nodejs bundle and 4.74KB in the web bundle.

```text
index.js → dist/microstates.cjs.js...
┌──────────────────────────────────────────┐
│                                          │
│   Destination: dist/microstates.cjs.js   │
│   Bundle Size:  22.55 KB                 │
│   Minified and Gzipped Size:  3.68 KB    │
│                                          │
└──────────────────────────────────────────┘

index.js → dist/microstates.es.js...
┌─────────────────────────────────────────┐
│                                         │
│   Destination: dist/microstates.es.js   │
│   Bundle Size:  34.46 KB                │
│   Minified and Gzipped Size:  4.74 KB   │
│                                         │
└─────────────────────────────────────────┘
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ With Microstates added to your project, you get:
 * 🎯 Transpilation free type system
 * 🔭 Optional integration with Observables
 * ⚛ Use in Node.js, browser or React Native
-* 🔬 Only [5.8kB gzipped](https://bundlephobia.com/result?p=microstates) with all dependencies
+* 🔬 Only [4.7kB gzipped](https://bundlephobia.com/result?p=microstates) with all dependencies
 
 But, most imporantly, Microstates makes working with state fun.
 


taras:
I'm using Bundlephobia as a reference for bundle size, but it doesn't know the size until release https://bundlephobia.com/result?p=microstates@0.10.1

taras:
It doesn't look like the bundlesize went down according to bundlephobia https://bundlephobia.com/result?p=microstates@0.11.0

cowboyd:
Yeah, I saw that. I think there are a couple of reasons.

1. Lenses are fully polymorphic using funcadelic type classes. I thought that we would need this feature, but we actually don't. It could be made a lot more simple just doing an `array` and `object` check.
2. We need to do an audit for dead code like this that's not called anywhere, or could be very easily removed https://github.com/microstates/microstates.js/blob/master/src/lens.js#L62-L77 probably worth something like ESLINT
3. The Query mechanism is now iterator based which is a bit more code, but will reduce the amount of application code, since there's no need to make custom types.

cowboyd:
Closing as it doesn't seem relevant. Can we just get a badge?

Issue #238, opened 2018-10-16
Make `ArrayType#map` function stable.


cowboyd:
One of the cornerstones of microstates is that setting a microstate to its existing value should result in the exact same microstate being returned. So, for example

```js
import { Any } from 'microstates';

let value = {x: 'y', z: 'w'}
let m = create(Any, value);
m === m.set(value).set(value).set(value) //=> true
```

The same should be true for an array microstate, and when using the `set` method it is.

```js
let value = [1, 2, 3];
let a = create([], value);
a === a.set(value).set(value).set(value) //=> true
```
☝️ This works today.

The same should hold for mapping an array. If the resulting mapped value is the same, then the mapped microstate should be _the_ same microstate (not an equivalent copy).

So, for example, if we were to map the identity function over the array microstate, then the value of the array doesn't change, and so the value of the array microstate doesn't change.

```js
let value = [1, 2, 3];
let a = create([], value);
a === a.map(x => x) //=> true
```
☝️ This does not currently work.

In fact, this is an expression of the [first functor law](https://en.wikibooks.org/wiki/Haskell/The_Functor_class#The_functor_laws).

By the same token, since microstates can be specified by their value, returning the same value for each constituent microstate should _also_ return the same original microstate, since the member at each index is the same, therefore the whole array value is the same, thereforce the array microstate should stay the same:

```js
import { valueOf } from 'microstates'

let value = [1, 2, 3];
let a = create([], value);
a === a.map(x => valueOf(x)) //=> true
```
☝️ This does not currently work.

or in the points free style:

```js
a === a.map(valueOf) //=> true
```

cowboyd:
closed by #239

PR #239, opened 2018-10-19
Make `ArrayType#map` function stable.

keathleydavidj:
Basically, when using the `.map` method on an Array microstate it was always returning a new instance of the Array even if the contents were the same.  Instead we expect that setting a microstate to its existing value should result in the exact same microstate being returned.
This PR intends to fix that, addressing #238 in the process.

*Before*
`a === a.map(x => x) //=> false`

*After*
`a === a.map(x => x) //=> true`

### DIFF
--- a/src/types/array.js
+++ b/src/types/array.js
@@ -1,5 +1,5 @@
 import { append } from 'funcadelic';
-import { Container, At } from '../lens';
+import { At, set } from '../lens';
 import { Profunctor, promap, mount, valueOf } from '../meta';
 import { create } from '../microstates';
 import parameterized from '../parameterized';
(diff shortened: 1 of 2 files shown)


src/types/array.js
```diff
@@ -48,9 +48,11 @@ export default parameterized(T => class ArrayType {
   }
 
   map(fn) {
-    return valueOf(this).map((item, index) => {
-      return valueOf(fn(create(T, item)));
-    });
+    let list = valueOf(this);
+    return list.reduce((acc, item, index) => {
```
cowboyd: I actually prefer not to use a generic variable like `acc` but instead use something with meaning in the context, like `values`.  Yes it is an accumulator in the context of an abstract reduction, but this is not an abstract reduction, it's a concrete reduction and the variable references very specific content: _the underlying array of values of the resulting microstate_.

PR #245, opened 2018-10-22
Introduce `ArrayType#sort` method

keathleydavidj:
This PR introduces the `.sort` method to Microstates' `ArrayType`.
It's basically a passthrough of `Array.prototype.sort` with the benefit that if the result of the `.sort()` call matches the original state, the original microstate is returned.

```
let sorted = create(ArrayType, ["a", "b", "c"]);
sorted === sorted.sort() //=> true
```

### DIFF
--- a/src/types/array.js
+++ b/src/types/array.js
@@ -41,6 +41,12 @@ export default parameterized(T => class ArrayType {
     return [value, ...valueOf(this)];
   }
 
+  sort(compareFn) {
+    let init = valueOf(this);
+    let result = [...init].sort(compareFn);
+    return init.every((member, idx) => result[idx] === member) ? this : result;
+  }
+
   filter(fn) {
     let list = valueOf(this);
     let result = list.filter((member) => fn(create(T, member)));
--- a/tests/types/array.test.js
+++ b/tests/types/array.test.js
@@ -59,6 +59,23 @@ describe("ArrayType", function() {
       });
     });
 
+    describe("sort", () => {
+      let sorted;
+
+      beforeEach(() => {
+        let unsorted = create([String], ["c", "b", "a"]);
+        sorted = unsorted.sort();
+      });
+
+      it("has a value that is the sorted version of the original value", () => {
+        expect(valueOf(sorted)).toEqual(["a", "b", "c"]);
+      });
+
+      it("returns the same array microstate if all of the values in the underlying array remains the same", () => {
+        expect(sorted.sort()).toBe(sorted);
+      });
+    });
+
     describe("filter", () => {
       let filtered;
 


tests/types/array.test.js
```diff
@@ -15,6 +15,7 @@ describe("ArrayType", function() {
 
     describe("push", () => {
       let pushed;
+
```
cowboyd: There are a lot of these extra spaces added, but as far as I can tell, they aren't related to the change. Part of good commit hygiene is to only include those things which pertain to what you're actually trying to accomplish. We can add this space and other in another PR (although should probably be accompanied by getting eslint set up)

tests/types/array.test.js
```diff
@@ -59,6 +61,23 @@ describe("ArrayType", function() {
       });
     });
 
+    describe("sort", () => {
+      let sorted;
+
+      beforeEach(() => {
+        let unsorted = create(ArrayType, [...valueOf(ms)].reverse());
```
cowboyd: There's no need to use the existing microstate. I think reversing it actually obfuscates the meaning of the test.

```js
let unsorted = create([String], ["c", "b", "a"]);
````
is much more clear.

File: src/types/array.js [hunk 3]
```diff
@@ -41,6 +41,18 @@ export default parameterized(T => class ArrayType {
     return [value, ...valueOf(this)];
   }
 
+  sort(compareFn) {
+    let init = valueOf(this);
+    let result = [...init].sort(compareFn);
+    let equalityCheck = (list1, list2) =>
```
cowboyd: Your equalityCheck is a fully general equality check, but there are several constraints here that let us optimize.

1) This is a sort, so the resulting list will always have the same number of members as the list it was derived from. As such, the length check isn't necessary.
2) If you're going to define this function inline, I'd just allow it to reference the `init` closure variable which is implicit. Otherwise, you would want to define the function outside of the method.
3) I'd recommend either a `for` loop or `Array.prototype.every` which allow you to return early because you know that two lists are not equal  _the first_ time you experience an inequality. With the reduction here, you're going to iterate through the entire list even if you see that the first two items are different:


Here's the `for` loop version:

```js
function didChange(result) {
  for (let i = 0; i < init.length; i++) {
    if (init[i] !== result[i]) {
      return false;
    }
  }
  return true;
}
```
See how it fails fast? But we should probably go with [some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) or [every](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) since they'd be more concise and functionally "hip"

tests/types/array.test.js
keathleydavidj: 👍 I've pulled these out for a separate ESLint PR.

src/types/array.js
keathleydavidj: 👍 I've switched to using `Array.prototype.every` here instead.

src/types/array.js
```diff
@@ -41,6 +41,16 @@ export default parameterized(T => class ArrayType {
...
   }
 
+  sort(compareFn) {
+    let init = valueOf(this);
+    let result = [...init].sort(compareFn);
+    let equalityCheck = comparatorList =>
+      init.every((member, index) =>
+        comparatorList[index] === member);
+
+    return equalityCheck(result) ? this : result;
```
cowboyd: Now that it's this concise, it should fit within 80 characters, so let's just inline it if we can..

tests/types/array.test.js
```diff
@@ -59,6 +59,23 @@ describe("ArrayType", function() {
...
 
+    describe("sort", () => {
+      let sorted;
+
+      beforeEach(() => {
+        let unsorted = create([String], ["c", "b", "a"]);
+        sorted = unsorted.sort();
+      });
+
+      it("has state", () => {
```
cowboyd: Let's make this a bit more descriptive:

"It has a value that is the sorted version of the original value"

PR #247, opened 2018-10-23
Update badges to put best foot forward.

### LINKED ISSUE #196: Coverage not accurate
The coveralls report appears to be quite stale

cowboyd:
The coverage level is inaccurate, and reflects poorly on the project. In fact, we're not running coverage reports at all, and if we were, I have to imagine that the coverage level would be higher than 71%.

This change removes the coverage badge entirely until we actually commit to checking code coverage.

It also add adds an npm version badge so that it's clear which version is the current one.

### Before
image

### After
image

resolves #196


taras:
Let me get the actual coverage report values - if I can.

PR #256, opened 2018-10-31
Fix a small code typo in the Readme

jeffutter:
I’m new to microstates and was just reading over the readme. I _think_ there is a small bug in the Array Microstates example.

The example creates a new `blog2` after immutably ‘pushing’ onto `blog.posts` but then logs out the values of `blog.posts` which wouldn’t have the newly pushed on value since `blog.posts` is immutable. I think the example should be logging out the new `blog2.posts`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,7 +299,7 @@ When you're working with an array microstate, the shape of the Microstate is det
 ```js
 let blog2 = blog.posts.push({ id: 3, title: "It is only getter better" });
 
-for (let post of blog.posts) {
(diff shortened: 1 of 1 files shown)


cowboyd:
Welcome to microstates @jeffutter! And how right you are. It's good to know that you were paying attention :)

As you noted, iterating over the original blog post would have yielded the same two items since it didn't change at all. 

Thanks again, and please let us know about any questions you might have about microstates.

PR #275, opened 2018-11-12
Added remove transition to ArrayType

taras:
This PR introduces a `remove` transition to ArrayType microstates to make it easier to "remove" elements from an array. It's using the filter transition which ensures that the same microstate is returned when there is a no-op operation.

Closes #275

### DIFF
--- a/src/types/array.js
+++ b/src/types/array.js
@@ -76,6 +76,10 @@ export default parameterized(T => class ArrayType {
     }, list);
   }
 
+  remove(item) {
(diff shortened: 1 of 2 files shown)


cowboyd:
Another thing that's interesting is that remove will work with either the microstate OR, t's value.

```js
let a = create(String, 'hi');
let array = create([String]).push(a).remove('hi').length //=> 0
```

taras:
That is interesting

cowboyd:
It feels somehow right though.

PR #276, opened 2018-11-12
Added object iterator to allow using ObjectType with queries

### LINKED ISSUE #224: Map for Object Microstates

taras:
This PR introduces `Symbol.iterator` for ObjectType. This makes it possible to use object microstates with `for (let obj of microstates)` syntax, which also provides support for Queries to map, filter and reduce objects.

The iterator yields entry objects which allow the key/value to be restructured as objects `for (let { key, value } of microstates)` or arrays `for (let [value key] of microstates)`.

Closes #224

### DIFF
--- a/src/query.js
+++ b/src/query.js
@@ -14,6 +14,17 @@ export function reduce(iterable, fn, initial) {
   return query(iterable).reduce(fn, initial);
 }
 
+export class Entry {
+  constructor(key, value) {
+    this.key = key;
+    this.value = value;
+  }
+
+  [Symbol.iterator]() {
+    return [this.value, this.key][Symbol.iterator]();
+  }
+}
+
 class Query {
   constructor(iterable) {
     if (typeof iterable[Symbol.iterator] === 'function') {
--- a/src/query.js
+++ b/src/query.js
@@ -14,17 +14,6 @@ export function reduce(iterable, fn, initial) {
   return query(iterable).reduce(fn, initial);
 }
 
-export class Entry {
-  constructor(key, value) {
-    this.key = key;
-    this.value = value;
-  }
-
-  [Symbol.iterator]() {
-    return [this.value, this.key][Symbol.iterator]();
-  }
-}
-
 class Query {
   constructor(iterable) {
     if (typeof iterable[Symbol.iterator] === 'function') {
--- a/src/types/object.js
+++ b/src/types/object.js
@@ -2,6 +2,7 @@ import { append, filter, map } from 'funcadelic';
 import parameterized from '../parameterized';
 import { mount, valueOf } from '../meta';
 import { create } from '../microstates';
+import { Entry } from '../query';
 
 export default parameterized(T => class ObjectType {
   static T = T;
@@ -52,7 +53,7 @@ export default parameterized(T => class ObjectType {
         let next = iterator.next();
         return {
           get done() { return next.done; },
-          get value() { return object[next.value]; }
+          get value() { return new Entry(next.value, object[next.value]) }
         };
       }
     };
--- a/src/types/object.js
+++ b/src/types/object.js
@@ -2,7 +2,6 @@ import { append, filter, map } from 'funcadelic';
 import parameterized from '../parameterized';
 import { mount, valueOf } from '../meta';
 import { create } from '../microstates';
-import { Entry } from '../query';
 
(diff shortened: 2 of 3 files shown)


cowboyd:
@rkeeler suggested in #276 (and my inclination is to agree) that we should be iterating over `key/value` pairs, not just values. In other words, an object is a stream of entries, not just values.

taras:
Yes, I was thinking the same thing. I just wasn't sure what was the best way to do this. Should it be `{key, value}`?

taras:
@cowboyd I made the changes as per our discussion in Slack

rkeeler:
I like that you can destructure the pairs in the for/of loop as either an array or an object. But I think the more common case over using for/of is using an anonymous function with the microstate queries. And in that case, my preference for the API would be making the key and value be separate arguments to the supplied function so that you don't have to deal with destructuring and can leave the key/index out most of the time. Granted it's a small sample size, but I've actually yet to need the key or index in any of the queries i've written. The majority look like this.

```js
get player() {
  return userName => find(this.players, p => p.userName.state === userName);
}
```

rkeeler:
Actually, looking into it a bit more. I think my preferred API for exposing the key/index to the query functions can still be achieved and isn't hindered by the implementation in this PR. It would be an additional change to the implementation of the query functions.

I had to add a few new queries for what i was doing (such as find and forEach) and I'll be making a PR with those soon. I could throw in changes to expose the keys and indexes to the queries in that PR.

src/query.js
```diff
@@ -14,6 +14,17 @@ export function reduce(iterable, fn, initial) {
...
 }
 
+export class Entry {
+  constructor(key, value) {
+    this.key = key;
+    this.value = value;
+  }
+
+  [Symbol.iterator]() {
+    return [this.value, this.key][Symbol.iterator]();
```
cowboyd: Sometimes you come across a line of code that is just pleasing to look at. This is one of those lines.

src/types/object.js
```diff
@@ -2,6 +2,7 @@ import { append, filter, map } from 'funcadelic';
 import parameterized from '../parameterized';
 import { mount, valueOf } from '../meta';
 import { create } from '../microstates';
+import { Entry } from '../query';
```
cowboyd: `Entry` is a concept of `Object`, and not `Query`, and so should probably reside in the object class.

cowboyd:
@rkeeler the Microstates query mechanism is actually drawn from a proof of concept that I through together here https://github.com/cowboyd/generation.js/blob/master/src/query.js

I'd like to eventually use this library instead since there's nothing really microstates about the query mechanism. It's really just comprehensions over iterables. 

`find` would make a welcome addition, and I'd also like to get `union`, `intersect`, `disjunct`, and friends in there as well.

src/query.js
taras: 🤗

taras:
@cowboyd I've made all of the changes.

cowboyd:
Can we make this a single merge commit?

PR #277, opened 2018-11-12
Failing test for transitioning microstates with initial value

minkimcello:
```js
  class Basic {
    bool = Boolean;
    numb = Number;
  }
  class Test {
    basic = create(Basic, { bool: false, numb: 5 });
  }
```

When creating a microstate with type Test, the created microstate loses initial value when transitioning composed microstate.

### DIFF
--- a/tests/microstates.test.js
+++ b/tests/microstates.test.js
@@ -5,8 +5,8 @@ import { valueOf } from '../src/meta';
 
 import Any from '../src/types/any';
 
-describe("Microstates", () => {
-  describe("default", () => {
+describe('Microstates', () => {
+  describe('default', () => {
     let def;
     beforeEach(function() {
       def = create();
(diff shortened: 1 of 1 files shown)


cowboyd:
Hi @minkimcello! And thanks for taking the time to report. This is an interesting corner case, and one that used to be handled by Microstates. 

We actually removed any cases of "looking down the tree" at child states and automatically incorporating them into the current value since it meant that microstates couldn't be lazy (in other words, you can't know your own value without materializing your child values). This also posed a problem for recursive microstates since you could very easy get into an infinite loop by looking at child values ad infinitum.

In the most recent microstates, every single transition ends up resolving, in effect,  to a lens set on the root value. In this case, the root value is `undefined`, so the transition in your test case results in the equivalent of a `set(lensPath(['basic', 'bool']), true, undefined)` which results in the next microstate value being `{ basic: { bool: true } }` which, now that it has a definition leaves out the value for `count`

Still, this is very counterintuitive, and I think we might be able to make a special case which would still allow microstates to be lazy and recursive while still being able to incorporate these into default values.

In the mean time, I would workaround this issue with an initializer:

```js
class Test {
  initialize(value = {}) {
    if (value.count == null || value.bool == null ) {
      return Object.assign({}, value, {
        get count() { return value.count == null ? 5 : value.count; }
        get bool() { return value.bool == null ? value.bool; }
      });
    } else {
      return value;
    }
  }
}
```

minkimcello:
Hello Charles, thank you so much for the thorough explanation. Working with microstates has been very exciting. :)

Taras also provided me with a workaround. What do you think?

```js
function beepBoop() {
  return class Test {
    bool = create(Boolean, false);
    count = create(Number, 0);
  };
}
```

taras:
@cowboyd do we need to create an issue out of this?

PR #284, opened 2018-11-23
Ensure Identity maps non-enumerable properties 

taras:
Our identity mechanism maps a microstate to create the identity. If the microstate's properties are not enumerable, then microstates at those properties don't get Identities created. This was causing a problem with Microstate created using `from` because object's properties were not getting wrapped in identities.

This PR changes the implementation of `promap` to use `Object.getOwnPropertyNames` instead of `Object.keys` because `Object.getOwnPropertyNames` includes non-enumerable properties.

I had to move existing tests into a describe blog to make debugging easier. The previous test setup was causing microstates to be created for every test which caused unnecessary `console.log` when I was trying to debug the problem.

### DIFF
--- a/src/identity.js
+++ b/src/identity.js
@@ -64,6 +64,8 @@ export default function Identity(microstate, observe = x => x) {
 
     let descriptors = Object.getOwnPropertyDescriptors(Type.prototype);
 
+    console.log(descriptors)
+
     let methods = Object.keys(methodsOf(Type)).concat(["set"]);
 
     Object.assign(Id.prototype, foldl((methods, name) => {
--- a/src/identity.js
+++ b/src/identity.js
@@ -64,8 +64,6 @@ export default function Identity(microstate, observe = x => x) {
 
     let descriptors = Object.getOwnPropertyDescriptors(Type.prototype);
 
-    console.log(descriptors)
-
     let methods = Object.keys(methodsOf(Type)).concat(["set"]);
 
     Object.assign(Id.prototype, foldl((methods, name) => {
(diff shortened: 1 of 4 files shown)


cowboyd:
I don't think this is the solution, because we only want to map those properties which are part of the core microstate graph,  and those that are merely incidental. For example if you have a computed property on a microstate class: 

```js
class TodoMVC {
  get latest() {
    return sort(this.todos, (a, b) => b.createdAt.state - a.createdAt.state).first
  }
}
```

We _don't_ want to promap this property, we actually want it to return the identity of the object where it is in the list of `this.todos` so that the object in the property is the _same_ as the object in the list.

Why not just make the properties enumerable in a literal?

taras:
> Why not just make the properties enumerable in a literal?

How do we do this? These properties are enumerable `from({ name: 'Taras' })`. The problem is that name property on microstate created with `from({ name: 'Taras' })` is not enumerable. That's why it doesn't get wrapped in Store. This problem happens with composed types as well.

cowboyd:
That's odd, I thought that our identity tests uses composed types via `TodoMVC`. 

For objects, we should be able to set the `enumerable: true` property to true when we define the descriptor for it https://github.com/microstates/microstates.js/blob/master/src/types/object.js#L15-L19

cowboyd:
This passed for me:

https://github.com/microstates/microstates.js/commit/f3e7093325996be1bd9f1a41e2da161cd2666da7

I'm still surprised that it isn't working correctly though because the `CachedProperty` that is used to define substates is, in fact, enumerable, and should be returned by `Object.keys` 

https://github.com/microstates/microstates.js/blob/cl%2Fenumerable-object-keys/src/cached-property.js#L2

taras:
@cowboyd do you want to create a PR or should I update this?

taras:
@cowboyd there is an ESLint error

PR #290, opened 2018-12-03
Gather transitions from prototype chain

### LINKED ISSUE #287: Store does not put closure around inherited transitions
When creating a microstate that extends another microstate via the `Store` the `this` context does not reference transition methods inherited from a parent.  

Given
```js
export default class Person {
  firstName = String;
  lastName = String;

  get fullName() {
    return `${this.firstName.state} ${this.lastName.state}`;
  }

  changeIdentity(firstName, lastName) {
    return this
      .firstName.set(firstName)
      .lastName.set(lastName);
  }
}

export default class Infant extends Person {
  get fullName() {
    return `Baby ${this.firstName.state} ${this.lastName.state}`;
  }
}

let infant = Store(create(Infant), nextState => this.set('infant', nextState));
```

When you try and call `infant.changeIdentity` the `this` context may reference a component's context instead of the microstates.

Reproduction of the failure can be seen here: https://github.com/brandynbennett/microstates-test

Also see discussion on https://github.com/microstates/ember/issues/78

kapture 2018-11-30 at 15 51 05

taras:
When a transition is inherited from the parent, the transition definition resides on the parent's prototype. This resulted in closures not being created around parent's transitions because `methodOf` function that was used to gather transitions was only looking on the class's prototype. 

This PR introduces `getPrototypeDescriptors` function that gathers prototypes by recursively walking the prototype chain. `methodOf` consumes this function to ensure that all transitions are included when creating closures for transitions.

Closes #287 
Related https://github.com/microstates/ember/issues/78

### DIFF
--- a/src/reflection.js
+++ b/src/reflection.js
@@ -1,7 +1,24 @@
 import { filter } from 'funcadelic';
 
+const { getPrototypeOf, getOwnPropertyDescriptors, assign } = Object;
+
 export function methodsOf(Type) {
   return filter(({ key: name, value: desc }) => {
     return name !== 'constructor' && typeof name === 'string' && typeof desc.value === 'function';
-  }, Object.getOwnPropertyDescriptors(Type.prototype));
+  }, getAllPropertyDescriptors(Type.prototype));
+}
+
+/**
+ * As opposed to `getOwnPropertyDescriptors` which only gets the
+ * descriptors on a single object, `getAllPropertydescriptors` walks
+ * the entire prototype chain starting at `prototype` and gather all
+ * descriptors that are accessible to this object.
+ */
+function getAllPropertyDescriptors(object) {
+  if (object === Object.prototype) {
+    return {};
+  } else {
+    let prototype = getPrototypeOf(object);
+    return assign(getAllPropertyDescriptors(prototype), getOwnPropertyDescriptors(object));
+  }
 }
--- a/tests/identity.test.js
+++ b/tests/identity.test.js
@@ -209,4 +209,25 @@ describe('Identity', () => {
       expect(i.name.state).toBe('Taras!!!');
     });
   });
+
+  describe('supports destructuring inherited transitions', () => {
+    class Parent {
+      a = String;
+      setA(str) {
+        return this.a.set(str);
+      }
+    }
+    class Child extends Parent {}
+
+    let i, setA;
+    beforeEach(() => {
+      i = Identity(create(Child), next => i = next);
+      setA = i.setA;
+    });
+
+    it('allows invoking a transition', () => {
+      setA('taras');
+      expect(i.a.state).toEqual('taras');
+    });
+  });
 });


src/reflection.js
```diff
@@ -1,7 +1,18 @@
...
+
 export function methodsOf(Type) {
   return filter(({ key: name, value: desc }) => {
     return name !== 'constructor' && typeof name === 'string' && typeof desc.value === 'function';
-  }, Object.getOwnPropertyDescriptors(Type.prototype));
+  }, getPrototypeDescriptors(Type));
 }
+
+export function getPrototypeDescriptors(constructor) {
+  let prototype = getPrototypeOf(constructor);
```
cowboyd: I'm actually confused how this works since `Type` is the constructor itself.

```js
class Foo {}
Object.getPrototypeOf(Foo) === Foo.prototype //=> false
Object.getPrototypeOf(Foo) === Function.prototype //=> true
```

taras: @cowboyd I pushed comment to the code to show you how it works.

src/reflection.js
```diff
@@ -1,7 +1,22 @@
...
   return filter(({ key: name, value: desc }) => {
     return name !== 'constructor' && typeof name === 'string' && typeof desc.value === 'function';
-  }, Object.getOwnPropertyDescriptors(Type.prototype));
+  }, getPrototypeDescriptors(Type));
 }
+
+export function getPrototypeDescriptors(Type) {
+  let prototype = getPrototypeOf(Type);
+  // do not include Object.prototype descriptors
+  if (prototype && prototype !== getPrototypeOf(Object)) {
```
cowboyd: Right, but:
```
> Object.getPrototypeOf(Object) === Object.prototype
false

> Object.getPrototypeOf(Object) === Function.prototype
true

```

In other words, are you sure `getPrototypeOf(Object)` and `getPrototypeOf(Type)` are doing what you think they're doing?

taras:
`prototype !== getPrototypeOf(Object)` there is probably a better way to express this but I'm checking if the type extends from another type, not from Object. IOW, I'm looking for cases where I need to recurse further,

```js
> Object.getPrototypeOf(Object) === Object.getPrototypeOf(class FooBar extends Foo {} )
false
```

cowboyd:
Ah, I see. That is actually a strange foible of the JavaScript runtime. Rather than relying on that, I think it would be clearer to make the recursion explicit:

```js
export function getPrototypeDescriptors(Type) {
  if (Type === Object) {
    return {};
  } else {
    let Super = getPrototypeOf(Type.prototype).constructor;
    return assign(getPrototypeDescriptors(Super), getOwnPropertyDescriptors(Type.prototype));
  }
}
```

cowboyd:
> Ah, I see. That is actually a strange foible of the JavaScript runtime. Rather than relying on that, I think it would be clearer to make the recursion explicit:

My apologies for this tone. I was rushing to get this feedback in between meetings and upon a re-read realized this might have come off as brusque and dismissive. In actuality, I actually learned something new yesterday which is that the prototype of a constructor function which extends another is not the same as `Function.prototype`, and there may be a very good reason for it (I just don't see one at the moment).

taras:
> My apologies for this tone. I was rushing to get this feedback in between meetings and upon a re-read realized this might have come off as brusque and dismissive.

Thanks but I didn't sense that. It's all good.

PR #302, opened 2018-12-17
Added Identity documentation

taras:
This addition was way over due. It describes two mechanisms for creating identities from microstates. 

[Rendered](https://github.com/microstates/microstates.js/blob/c6946554ce6b65a2683704e24a97c129ea208d90/README.md#streaming-state)

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,13 @@ feel like working with the classic, mutable models we all know and love.
   - [create(Type, value): Microstate](#createtype-value-microstate)
   - [from(any): Microstate](#fromany-microstate)
   - [map(microstate, fn): Microstate](#mapmicrostate-fn-microstate)
-- [Observable Microstates](#observable-microstates)
+- [Streaming State](#streaming-state)
+  - [Structural Sharing](#structural-sharing)
+  - [Memoized Getters](#memoized-getters)
+  - [Debounce no-op transitions](#debounce-no-op-transitions)
+  - [Identity Constructors](#identity-constructors)
+    - [Store(microstate, callback)](#storemicrostate-callback)
+    - [Observable.from(microstate)](#observablefrommicrostate)
 - [The Vision of Microstates](#the-vision-of-microstates)
   - [Shared Solutions](#shared-solutions)
   - [Added Flexibility](#added-flexibility)
@@ -128,13 +134,13 @@ A Microstate is just an object that is created from a value and a type. The valu
 Microstates comes out of the box with 5 primitive types: `Boolean`, `Number`, `String`, `Object` and `Array`.
 
 ```js
-import { create, valueOf } from "microstates";
+import { create, valueOf } from 'microstates';
 
 let meaningOfLifeAndEverything = create(Number, 42);
 console.log(meaningOfLifeAndEverything.state);
 //> 42
 
-let greeting = create(String, "Hello World");
+let greeting = create(String, 'Hello World');
 console.log(greeting.state);
 //> Hello World
 
@@ -143,7 +149,7 @@ console.log(isOpen.state);
 //> true
 
 // For Object and Array use microstates valueOf method
-let foo = create(Object, { foo: "bar" });
+let foo = create(Object, { foo: 'bar' });
 console.log(valueOf(foo));
 //> { foo: 'bar' }
 
@@ -164,9 +170,9 @@ class Person {
 Once you have a type, you can use that type to create as many people as your application requires:
 
 ```js
-import { create } from "microstates";
+import { create } from 'microstates';
 
-let person = create(Person, { name: "Homer", age: 39 });
+let person = create(Person, { name: 'Homer', age: 39 });
 ```
 
 Every microstate created with a type of `Person` will be an object
@@ -207,8 +213,8 @@ class Car {
 }
 
 let theHomerCar = create(Car, {
-  designer: { name: "Homer", age: 39 },
-  name: "The Homer"
+  designer: { name: 'Homer', age: 39 },
+  name: 'The Homer'
 });
 ```
 
@@ -261,7 +267,7 @@ You can use the `valueOf()` function available from the microstates
 module to retrieve the underlying value represented by a microstate.
 
 ```js
-import { valueOf } from "microstates";
+import { valueOf } from 'microstates';
 
 valueOf(theHomerCar);
 //> { designer: { name: 'Homer', age: 39 }, name: 'The Homer' }
@@ -283,8 +289,8 @@ class Post {
 
 let blog = create(Blog, {
   posts: [
-    { id: 1, title: "Hello World" },
-    { id: 2, title: "Most fascinating blog in the world" }
+    { id: 1, title: 'Hello World' },
+    { id: 2, title: 'Most fascinating blog in the world' }
   ]
 });
 
@@ -298,7 +304,7 @@ for (let post of blog.posts) {
 When you're working with an array microstate, the shape of the Microstate is determined by the value. In this case, `posts` is created with two items which will, in turn, create a Microstate with two items. Each item will be a Microstate of type `Post`. If you push another item onto the `posts` Microstate, it'll be treated as a `Post`.
 
 ```js
-let blog2 = blog.posts.push({ id: 3, title: "It is only getter better" });
+let blog2 = blog.posts.push({ id: 3, title: 'It is only getter better' });
 
 for (let post of blog2.posts) {
   console.log(post);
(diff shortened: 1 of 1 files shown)


taras:
@alexlafroscia can you please read this and tell me if it makes sense to you?

README.md
```diff
@@ -771,28 +777,85 @@ let numbers = create([Number], [1, 2, 3, 4]);
...
-receive a microstate with middleware installed that will cause the result of transitions to be pushed
-through the stream.
+To bring a reactive system to life, we need to capture the latest state in a sequence of states. In component driven reactive engines like React, Vue, Ember & Angular the state needs to emerge at the root of the component tree. When a new state is emitted, we need to apply the state to the root component and allow the state to cascade down the component tree.
 
-You should be able to use to any implementation of Observables that supports `Observer.from` using [symbol-observable](https://github.com/benlesh/symbol-observable). We'll use `RxJS` for our example.
+Microstates has a mechanism to working with Microtates in reactive environments. It is called Identity. When you create an Identity from a Microstate, you get an object that has the same shape the original microstate. Every composed microstate becomes an identity and every transition gets wrapped in side effect emitting behaviour specific to that identity constructor. Let’s look at the features of the Identity and how they benefit reactive environments.
```
alexlafroscia:
```suggestion
Microstates has a mechanism for working with reactive environments: Identity. When you create an Identity from a Microstate, you get an object that has the same shape the original microstate. Let’s look at the features of the Identity and how they benefit reactive environments.
```

I think there was a little too much context given up front there. Better to cut it down a bit and let the example explain how it's used.

README.md
```diff
@@ -771,28 +777,85 @@ let numbers = create([Number], [1, 2, 3, 4]);
...
+
+## Structural Sharing
+
+A common performance optimization used by all reactive engines is to prevent re-renders for components who’s props have not changed. The most efficient way to determine if a value has not changed it to perform an exact equality check, for example: `prevValue === currentValue`. If the reference is the same, then consider the value unchanged. The Identity makes this possible with Microstates by internally managing how the Identity is constructed as a result of a transition. It will automatically determine which branches of microstates are unchanged and reuse previous identities for those branches.
+
+## Memoized Getters
+
+Microstates are immutable which makes it safe for us to memoize computations that are derived off their state. Identies will automatically memoize getters and return previously computed value when the microstate backing the identity has not changed. When a transition is invoked on the identity, the part of the identity tree that are changed will be re-created effectively invalidating the cache for the changed parts of the identity. The getters will be recomputed for state that is changed.
+
+## Debounce no-op transitions
```
alexlafroscia: How is a no-op calculated? By which I mean to ask, a transition from and back to the original state within what time period avoids emitting a new Identity? The following code performs two transitions

```javascript
let id = Store(from({ foo: 'bar' }), prev => {
  console.count('changed');
  id = prev;
});

id.foo.set('bax');
id.foo.set('bar');
```

README.md
```diff
@@ -771,28 +777,85 @@ let numbers = create([Number], [1, 2, 3, 4]);
...
+
+Identities automatically prevent unnecessary re-renders by debouncing transitions that do not change the value. This eliminates the need for `shouldComponentUpdate` hooks for pure components because it is safe to assume that a component that is re-rendering is re-rendering as a result of a transition that changed the value.
+
+## Identity Constructors
+
+Microstates comes with two Identity constructors: _Store_ and _Observable_. Observable will create a stream of identities. Next identity will be sent to the observer when an transition on an identity is called.
```
alexlafroscia: Just for symmetry, if you're going to give a basic explanation of what `Observable` does in this instance, I think it's also worth giving a short one-sentence description of what `Store` does as well.

README.md
```diff
@@ -771,28 +777,85 @@ let numbers = create([Number], [1, 2, 3, 4]);
...
+Store identity constructor takes two arguments: microstate and a callback. It returns an identity. When a transition is invoked on the identity, the callback will receive the next identity.
+
+```js
+import { Store, from, valueOf } from 'microstates';
+
+let initial = create(Number, 42);
+
+let last;
+
+last = Store(initial, next => (last = next));
```
alexlafroscia: Does the `Store` callback get invoked for the initialize transition? Or just every future transition?

README.md
```diff
@@ -771,28 +777,85 @@ let numbers = create([Number], [1, 2, 3, 4]);
...
+### Observable.from(microstate)
+
+Microstates provides an easy way to convert a Microstate which represents a single value into a Observable stream of values. This is done by passing a Microstate to Observable.from function. This function will return a Observable object with a subscribe method. You can subscribe to the stream by passing an observer to the subscribe function. Once you subscribe, you will synchronously receive a microstate with middleware installed that will cause the result of transitions to be pushed through the stream.
+
+You should be able to use to any implementation of Observables that supports Observer.from using symbol-observable. We'll use RxJS for our example.
 
 ```js
-import { from } from "rxjs";
-import { create } from "microstates";
+import { from } from 'rxjs';
```
alexlafroscia: Not sure if you want to do this in the documentation, but in my own code that uses Microstates and RXJS I did this:

```javascript
import { from as createObservable } from 'rxjs';
```

just because `microstates` _also_ has a `from` function and I wanted to make it more clear which was being used at a glance.

README.md
```diff
@@ -771,28 +777,85 @@ let numbers = create([Number], [1, 2, 3, 4]);
...
+  render() {
+    let { last } = this.state;
+    return (
+      <button onClick={() => last.increment()}>Increment {last.state}</button>
+    );
+  }
+}
+```
+
+### Observable.from(microstate)
```
alexlafroscia: I think that it might be useful to talk about when you might need to "upgrade" from using `Store` to a full-blown Observable library. This isn't really clear to me; from my perspective it feels like there are just two ways to do pretty much the same thing, but Observable uses an existing concept and is more powerful, as you also get all the benefits of the rest of whichever library you're using.

While `Store` has a very compelling use-case, like the React example above shows, it feels kind of weird to me that there are two different ways to subscribe to state changes. I'd almost rather there just be one, to cut down on API surface area and remove a layer of decisions that need to be made.

alexlafroscia:
I have a few more general questions about Store and Identity:

- Why is the public function called `Store`?
- Why is that different from the internal function, `Identity`?
- Why introduce an additional concept, `Identity`, when it doesn't feel that different than `Microstate`? Asked another way; should `from` and `create` also return an `Identity` instead of a `Microstate`?

README.md
taras: > How is a no-op calculated?

It's based on value equality. Using your example,

```js
id.foo.set('bax');
id.foo.set('bax');
```

Would only cause one update because 2nd operation is debounced.

alexlafroscia: Oh okay, I see what you're saying -- if the current state and the state being transitioned to are the same, then a "new state" that would be the same _is not_ emitted.

README.md
alexlafroscia: To answer my own question, it seems like it's only for future transitions, so you need to use the return value from `Store` for the initial state of the `last` variable. The following _does not_ work:

```
let last;

Store(initial, next => (last = next));

last.foo.set('bax');
```

README.md
taras: Yes.

README.md
```diff
@@ -771,28 +777,81 @@ let numbers = create([Number], [1, 2, 3, 4]);
...
 
-# Observable Microstates
+# Streaming State
 
-By themselves microstates are purely functional. They have no builtin concept of identity,
-time or sequencing. However, it is precisely because of these properties that they can be
-used with almost any state management solution that does. In fact, Microstates comes bundled
-out of the box with the ability to convert any microstate into a stream of transitions using
-the the Observable API.
+A microstate represents a single immutable value and provides transitions to derive the next value. Microstates provides a mechanism called Identity that allows to emit stream of Microstates. When you create an Identity from a Microstate, you get an object that has the same shape the original microstate. Every composed microstate becomes an identity and every transition gets wrapped in side effect emitting behaviour specific to the identity's constructor. Identity wrapped Microstate offer the following benefits over raw Microstates.
```
alexlafroscia:
```suggestion
A microstate represents a single immutable value with transitions to derive the next value. Through a mechanism called `Identity`, you can emit stream of Microstates as the value is updated over time.

When you create an Identity from a Microstate, you get an object that has the same shape the original microstate. Every composed microstate becomes an identity and every transition gets wrapped in side effect emitting behavior specific to the identity's constructor. Identity wrapped Microstate offer the following benefits over raw Microstates.
```

alexlafroscia: This seems _way_ more clear to me now. Basically, `Identity` is the original `Microstate` plus the additional logic that allows you to subscribe to the changes over time, either through `Store` or `Observable`. Is that right?

README.md
taras: I'll do this: `import { from as observableFrom } from 'rxjs'`

README.md
taras: Observable interface was designed specifically for Observable interop. It's not really meant to be used over Store, unless you're using Observables. Internally, Observable interop using `Identity` which is the same thing as a Store. 

I'll answer the rest of this the other message with questions.

README.md
taras: Yes

README.md
taras: Yes, initialize transition lazily sets up the microstate on create. When you construct the Microstate, the value is not initialized until you still pulling values off the microstate. If your microstate is bound to the DOM, the render will pull the values off the microstate which will invoke initialize but it's a pure operation without side effects.

taras:
> Why is the public function called Store?

I wanted to keep it familiar for people, but Identity's capabilities keep growing. 

> Why is that different from the internal function, Identity?

They're exactly same right now. In the future, it might be helpful to make Identity the name for streams of microstates. Identity is become much more powerful than Store from other libraries. For example, in the References work which is currently in [microstates/lab](https://github.com/microstates/lab), there is a change to the identity that will make integration with Redux basically 3 lines of code. Once we land References, we'll make it possible for the developer to subscribe to state transitions on identity leafs. 

For example, you'll be able to do this:

```js
class App {
  session = Authentication;
}

let id;
id = Store(create(App), next => id = next);

Observable.from(id.session)
  .subscribe(session => console.log('session state', session))
```

These are very special attributes of Identity that are not present in other frameworks. Before Microstates 1.0, we'll need to button up how we describe Identities and finalize the APIs. 

> Why introduce an additional concept, Identity, when it doesn't feel that different than Microstate?

The idea is that Microstates is the underlying immutable state primitive without side effects. Identity adds side effects and concerns around structural sharing and caching. By keeping this concepts separate, people will be able to compose them as necessary. 

> Asked another way; should from and create also return an Identity instead of a Microstate?

I don't think so. It's kinda how Immutable.js can be used with Redux but conflating them makes both concepts more limited.

PR #308, opened 2019-01-08
Fix minor README.md typo

leonardodino:
I'm not a user, yet, of this library, but it presents a nice concept for handling composition of state.

Just saw a minor syntax error while reading through (:

Nice Work!


taras:
👍 thank you

PR #312, opened 2019-01-25
[WIP] - Represent every microstate internally with an object.

cowboyd:
JavaScript has the concept of "primitive" types and "object" types. Number literals like `5` and string literals like "five" are not actually objects even though they have object methods available to them. Unlike an object, they are _sealed and immutable_. You may not change, add, or remove any of their properties. 

The reason this is so, and why it's important to microstates, is that they do not create a separate identity for each usage of a primitive. No matter how it is computed, either as a literal, or derived from some other value, each value of a primitive is indistinguishable from another usage of that same value. So, for example, `2+3 === 5` will always be true since what is being compared with strict equality is the in-memory representation of the number five. Contrast this with object types where the in-memory representation is effectively a pointer to the actual object, so that address is what gives the object its exclusive identity. In other words, `new Number(5) === new Number(5)` is false because each object has a different address.

### Why is this a problem for microstates?

This presents a challenge because there's no convenient way to uniquely identify a microstate which is something that we have to do if we want to make the elements in a store stable. The low-level microstates are pure functions, and so each microstate represents a unique point in time, so we cannot use the reference to the microstate itself.

Currently, what we do is do a convoluted tree-mapping excercise where we use the triple of (Type, value, path) to uniquely identify a microstate.  But this has some drawbacks.

1. It's awkward and the functional mapping of the tree is difficult to understand. Contrast this with a simple map lookup that we could do with a unique value.
2. As a follow on to the above, the only way to "lookup" a stable microstate in the store is by walking to its path from the root. Using a unique object, we can store that information directly in a weakmap or even on the object itself.
3. The memory implications are unknown. Even with the current implementation, memory leaks of the stable proxy objects could be kept around for microstates that are no longer in the tree.

If on the other hand, every microstate value was an object, we could store the stable proxies in a weak map, so that they would be trivial to look up, and the lookup could be done without walking the microstate tree to the point where the proxy lived.

### Approach

This change attempts to enable just a weakmap based proxy system by ensuring that every value in the value tree of a microstate is represented by an object, not a value. When you call `create(Type, value)`, `value` will be "upgraded" to an appropriate object. Therefore inside microstate transitions you are _always_ dealing with an object. primitive strings become `String` objects, primitive numbers become `Number` objects, and `null` and `undefined` are upgraded to instances of the classes and `Undefined` respectively.

In other words:

```js
valueOf(create(Number, 5)) //=> Number { 5 }
valueOf(create(Number, 5)).valueOf() //=> 5 
valueOf(create(Any, null)) //=> Null { }
valueOf(create(Any, null)).valueOf() //=> null
```

### Concerns / Questions

- This slightly changes the meaning of `valueOf` even if the contract remains the same since it really isn't a thing that you can use for equality comparison anymore. For example, the `Array#filter` method had to be re-written as such:

```js
remove(item) {
    return this.filter(s => valueOf(s).valueOf() !== valueOf(item).valueOf());
  }
```

notice the `valueOf(x).valueOf()`. This is becomes much more prevalent such that it makes me think that we might need another helper like `stateOf(x)` that is equilavent to `valueOf(x).valueOf()`

- Since a microstate should never have anything but an object value, do we want to warn if initialize returns something other than that, or again, just silently upgrade it.

- Why allow `null` or `undefined` at all? Instead, it seems more valuable to have an "empty" function which returns the empty value, `{}` for Object, `[]` for Array, '' for String, and default of `{}` for everything else.

- *Red-hydrating*: Primitive wrappers like `Number` serialize via JSON to their primitive counterparts, which makes this change particularly low-impact. In other words, `JSON.stringify(new Number(5))` => `"5"`. But what do we do when we want to go in the other direction?

```js
create(ArrayType.of(Num), [1, 2, 3]))
```

We want to be able to create proxies out of this from the get-go, but how do we do it short of a full mapping of the Array into its `Number` object equivalents before it can be used? And how can we efficiently do this for the entire tree? Can it be done lazily? I don't know.

### DIFF
--- a/src/identity.js
+++ b/src/identity.js
@@ -102,6 +102,6 @@ function equals(id, other) {
   if (other == null) {
     return false;
   } else {
-    return other.constructor.Type === id.constructor.Type && valueOf(id) === valueOf(other);
+    return other.constructor.Type === id.constructor.Type && valueOf(id).valueOf() === valueOf(other).valueOf();
   }
 }
--- a/src/lens.js
+++ b/src/lens.js
@@ -63,7 +63,7 @@ export function At(property, container = {}) {
   let get = context => context != null ? context[property] : undefined;
   let set = (part, whole) => {
     let context = whole == null ? (Array.isArray(container) ? [] : {}) : whole;
-    if (part === context[property]) {
+    if (primitiveOf(part) === primitiveOf(context[property])) {
       return context;
     } else if (Array.isArray(context)) {
       let clone = context.slice();
@@ -79,4 +79,9 @@ export function At(property, container = {}) {
 
 export function Path(path = []) {
   return path.reduce((lens, key) => compose(lens, At(key)), transparent);
+}
+
+
+function primitiveOf(object) {
+  return object == null ? object : object.valueOf();
 }
--- a/src/literal.js
+++ b/src/literal.js
@@ -2,10 +2,10 @@ import { create } from './microstates';
 
 class Literal {
   initialize(value) {
+    value = value.valueOf();
     if (value == null) {
       return this;
     }
-    value = value.valueOf();
     switch (typeof value) {
       case "number":
         return create(Number, value);
--- a/src/meta.js
+++ b/src/meta.js
@@ -1,4 +1,5 @@
 import { type, append } from 'funcadelic';
+import objectOf from './object-of';
 import { At, compose, transparent, over, view } from './lens';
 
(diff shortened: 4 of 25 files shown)


cowboyd:
I'm closing this for now since we were able to implement a structured path map without needing to weakly store proxies by value. I can see this still being something we could do in the future should the need arise, but otherwise it will always be here in a branch.

PR #314, opened 2019-02-02
Remove dead property caching code.

cowboyd:
In Ids, we were permanently caching computed properties, but none of the tests actually predicated on this. We may want to add this back in the future, but then again, maybe not.

If a computed property operates on a cached value, like a query, then while the query will not be stable, the objects contained in the query will. This is why the tests are passing, and why this can safely be removed for now.


cowboyd:
How is it possible to remove code and decrease coverage? 🤔

PR #319, opened 2019-02-13
Base the identity store on a map of paths to stable references.

cowboyd:
The store implementation is currently based on a tree mapping algorithm that, though lazy, is theoretically a mapping of the entire microstate tree. This worked well when what we were doing with the store was walking the entire tree as part of a render() cycle that starts at the top and then enumerates all children.

The problem with this is that it does not allow you to enter the graph at any point. This is pretty necessary when modelling side-effects, where you might want to be able to find out what the current stable reference is for a node that you might be working with over a long time.

Originally modelled using This replaces the tree mapping strategy with a "pathmap" strategy. When a microstate is enumerated or accessed, a `Location` object is allocated for it at that path. So, for example the progress of the second upload of the uploader would have a path `['uploads', 1, 'progress']`. Once allocated, that location _never changes_, which allows us to look up a location directly if we have its path.

The `Location` has all the information needed to work with that path: the current POJO value at that path, the microstate in the tree at that path, and the `reference` which is of the same type as the microstate that will be stable and proxy that microstate.

While the `Location` never changes, the `reference` that it holds does change as the underlying value changes. This is what provides us with a stable, yet immutable structure.

So, in our uploader example:

```js
class Uploader {
  uploads = [class Upload {}]
}
```

The location / reference structure would look like:

```
┌───────────┐                    ┌───────────┐
│           │                    │           │
│ Location  │───reference───────▶│ Uploader  │
│           │                    │           │
└──────┬────┘                    └─────┬─────┘
       │                               │
       │                               │
       │                               │          ┌───────────┐
       │         ┌───────────┐         │ uploads  │           │
       │uploads  │           │         └─────────▶│ [Upload]  │
       └────────▶│ Location  │──reference────────▶│           │
                 │           │                    └───────────┘
                 └────┬──────┘                         │
                      │                                            ┌────────┐
                      │           ┌───────────┐        │           │        │
                      │ 0         │           │        │─ ─ ─ ─ ─ ▶│ Upload │
                      ├──────────▶│ Location  │───reference───────▶│        │
                      │           │           │        │           └────────┘
                      │           └───────────┘
                      │                                │
                      │                                            ┌────────┐
                      │           ┌───────────┐        │           │        │
                      │ 1         │           │         ─ ─ ─ ─ ─ ▶│ Upload │
                      └──────────▶│ Location  │───reference───────▶│        │
                                  │           │                    └────────┘
                                  └───────────┘
```
Notice how the array of Uploads `[Upload]` does not actually have keyed access to its children. Those are lazily enumerated. However, once they _are_ enumerated, then we can assign them a location based on the position they appeared in when they were enumerated. This allows us to leap right back to that position if we need to work with this element in the future (as in the case where we want to report progress on a specific upload as it happens).

Notes:

 - I'm not thrilled about the `defineChildren` mechanism, but there has to be a way for these references to "intercept" children and replace them with further references. Specifically, this behavior must be different between Arrays and other container classes. I think this can be generalized in the future, but I think that there has to be some clarification on where the DSL officially hooks into the relationship definition process. I think

 - The `Profunctor` typeclass is now completely obsolete and should be removed in a follow-on change.

 - The storage class is just a very light-weight method for keeping a POJO, working with various paths within it, and receiving notifications of when it changes.

### DIFF
--- a/tests/pathmap.test.js
+++ b/tests/pathmap.test.js
@@ -37,7 +37,7 @@ describe('pathmap', ()=> {
     expect(valueOf(id)).toBe(ref.get());
   });
 
-  it('has children corresponding to the shit to the substates', ()=> {
+  it('has children corresponding to the substates', ()=> {
     expect(id.hall).toBeInstanceOf(BooleanType);
     expect(id.closet).toBeInstanceOf(BooleanType);
   });
--- a/tests/pathmap.test.js
+++ b/tests/pathmap.test.js
@@ -1,0 +1,123 @@
+/* global describe, beforeEach, it */
+
+import expect from 'expect';
+
+import Pathmap from '../src/pathmap';
+import { valueOf, BooleanType, ArrayType, NumberType } from '../index';
+
+describe('pathmap', ()=> {
+
+  let pathmap;
+  let LightSwitch;
+  let ref;
+  let id;
+  let hall;
+  let closet;
+
+  let current;
+  beforeEach(()=> {
+    ref = Ref({});
+    LightSwitch = class LightSwitch {
+      hall = Boolean;
+      closet = Boolean;
+    };
+    pathmap = Pathmap(LightSwitch, ref);
+    id = pathmap.get();
+    hall = id.hall;
+    closet = id.closet;
+
+    current = pathmap.get;
+  });
+
+  it('exists', () => {
+    expect(pathmap).toBeDefined();
+  });
+  it('has an id delegate which is represents the microstate at the base path', ()=> {
+    expect(id).toBeInstanceOf(LightSwitch);
+    expect(valueOf(id)).toBe(ref.get());
+  });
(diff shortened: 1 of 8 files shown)


tests/pathmap.test.js
```diff
@@ -0,0 +1,123 @@
...
+
+  it('exists', () => {
+    expect(pathmap).toBeDefined();
+  });
+  it('has an id delegate which is represents the microstate at the base path', ()=> {
+    expect(id).toBeInstanceOf(LightSwitch);
+    expect(valueOf(id)).toBe(ref.get());
+  });
+
+  it('has children corresponding to the shit to the substates', ()=> {
```
taras: `shit` might not be the right word there.

cowboyd: oops :)

PR #323, opened 2019-02-17
Don't try to reify a value if the iteration is finished.

cowboyd:
For each `step` of an iteration, the iterator has two properties: `step.done` and `step.value`. In the case of arrays and objects, if the `step.done` method returns false, the expectation is that you've moved past the end of list collection and that you won't even bother with calling the `step.value` because why? And this is mostly true of libraries that use iterator.

However, in the case of generator functions, each `step` represents the invocation of `yield`, and the _last_ step, in which `step.done` is `true` represents the return value of the generator function. In other words, `step.value` _can_ have a meaningful value. So for libraries that generalize over all iterators, not just collection iterators (like regenerator-runtime), we need to have `value` return something meaningful, not just a garbage entry with an invalid value.

Luckily, this is straightforward. For both `ObjectType`, and `ArrayType` we guard the iteration step value by checking first to see if the iteration is done, and return `undefined` if we've moved past the end of the collection.

### DIFF
--- a/src/types/array.js
+++ b/src/types/array.js
@@ -86,7 +86,13 @@ export default parameterized(T => class ArrayType {
         let index = i++;
         return {
           get done() { return next.done; },
-          get value() { return !next.done ? childAt(index, array) : undefined; }
+          get value() {
+            if (!next.done) {
+              return childAt(index, array);
+            } else {
+              return undefined;
+            }
+          }
         };
       }
     };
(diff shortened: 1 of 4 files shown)


cowboyd:
It should be noted that these types of guards would be unnecessary if we were to use generator function inside microstates proper to express the collection iterators. However, we've held off doing so in order to avoid using `regenerator-runtime`

PR #338, opened 2019-03-22
Don't store keys on the object type.

cowboyd:
Objects are a bit weird in that they actually keep all of the keys as properties of the object microstate. This has a couple of problems:

Performance
====

Because we need to iterate over all of the keys of the object value and define a property for them, it means a significant performance hit. Even though the values themselves are built lazily, the property definitions need to be created on every single construction.

Naming conflicts
====

Let's say you have an object that has keys with same names as transitions, like `{ put: 'the plate on the table' }`. This means that the _property_ "put" will override the _transition_ "put".

This seems untenable long term since it applies to every transition, including `initialize`, `delete`, `filter`..... all of them. The microstate would most likely break completely if the value had a property called "initialize".

This removes the direct properties from the Object microstate. They are still available as the `entries` property, but now they are completely contained within their own namespace and can't interfere with the Object namespace.

In practice, I expect that the `entries` property will be fairly rarely used, and in actual application, the way developers will access the substates in an object will be through enumeration.

### DIFF
--- a/src/types/object.js
+++ b/src/types/object.js
@@ -1,25 +1,20 @@
 import { append, filter, map } from 'funcadelic';
+import { reduce } from '../query';
 import parameterized from '../parameterized';
-import { valueOf } from '../meta';
+import { valueOf, mount } from '../meta';
 import { create } from '../microstates';
+import { Tree, childAt } from '../tree';
+
 
 export default parameterized(T => class ObjectType {
   static T = T;
 
-  static get name() {
-    return `Object<${T.name}>`;
-  }
+  static name = `Object<${T.name}>`
 
-  constructor(value) {
-    Object.keys(value || {}).forEach(key => {
-      Object.defineProperty(this, key, {
-        enumerable: true,
-        configurable: true,
-        get() {
-          return create(T, value[key]);
-        }
-      });
-    });
+  get entries() {
+    return reduce(this, (entries, entry) => Object.assign(entries, {
+      [entry.key]: entry.value
+    }), {});
   }
 
   initialize(value) {
@@ -56,7 +51,7 @@ export default parameterized(T => class ObjectType {
           get done() { return next.done; },
           get value() {
             if (!next.done) {
-              return new Entry(next.value, object[next.value]);
+              return new Entry(next.value, childAt(next.value, object));
             } else {
               return undefined;
             }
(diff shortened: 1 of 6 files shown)


cowboyd:
For some context:

### Before

```
┌────────┬──────────────────────────────────┬─────────┬──────────┬─────────────┐
│ Suite  │ Benchmark                        │ ops/sec │ σ        │ sample size │
├────────┼──────────────────────────────────┼─────────┼──────────┼─────────────┤
│ Object │ create({})                       │ 194,750 │ ± 1.12 % │ 87          │
├────────┼──────────────────────────────────┼─────────┼──────────┼─────────────┤
│ Object │ create({}).set({})               │ 110,619 │ ± 0.71 % │ 89          │
├────────┼──────────────────────────────────┼─────────┼──────────┼─────────────┤
│ Object │ create({}, BIGOBJECT)            │ 2,234   │ ± 2.28 % │ 83          │
├────────┼──────────────────────────────────┼─────────┼──────────┼─────────────┤
│ Object │ create({}).set(BIGOBJECT)        │ 2,224   │ ± 2.10 % │ 85          │
├────────┼──────────────────────────────────┼─────────┼──────────┼─────────────┤
│ Object │ Store(create({}))                │ 7,266   │ ± 3.62 % │ 80          │
├────────┼──────────────────────────────────┼─────────┼──────────┼─────────────┤
│ Object │ Store(create({}, BIGOBJECT))     │ 451     │ ± 2.49 % │ 77          │
├────────┼──────────────────────────────────┼─────────┼──────────┼─────────────┤
│ Object │ Store(create({}).set(BIGOBJECT)) │ 318     │ ± 2.22 % │ 83          │
└────────┴──────────────────────────────────┴─────────┴──────────┴─────────────┘
```

### After

```
┌────────┬──────────────────────────────────┬─────────┬──────────┬─────────────┐
│ Suite  │ Benchmark                        │ ops/sec │ σ        │ sample size │
├────────┼──────────────────────────────────┼─────────┼──────────┼─────────────┤
│ Object │ create({})                       │ 190,990 │ ± 0.94 % │ 85          │
├────────┼──────────────────────────────────┼─────────┼──────────┼─────────────┤
│ Object │ create({}).set({})               │ 104,149 │ ± 1.64 % │ 86          │
├────────┼──────────────────────────────────┼─────────┼──────────┼─────────────┤
│ Object │ create({}, BIGOBJECT)            │ 355,528 │ ± 0.83 % │ 90          │
├────────┼──────────────────────────────────┼─────────┼──────────┼─────────────┤
│ Object │ create({}).set(BIGOBJECT)        │ 101,192 │ ± 0.87 % │ 90          │
├────────┼──────────────────────────────────┼─────────┼──────────┼─────────────┤
│ Object │ Store(create({}))                │ 6,836   │ ± 3.78 % │ 77          │
├────────┼──────────────────────────────────┼─────────┼──────────┼─────────────┤
│ Object │ Store(create({}, BIGOBJECT))     │ 6,992   │ ± 3.32 % │ 81          │
├────────┼──────────────────────────────────┼─────────┼──────────┼─────────────┤
│ Object │ Store(create({}).set(BIGOBJECT)) │ 5,215   │ ± 2.93 % │ 83          │
└────────┴──────────────────────────────────┴─────────┴──────────┴─────────────┘
```

taras:
wow, that's a huge difference

cowboyd:
Yeah, over 10x improvement. I was suprised it was that much, but `defineProperty` is actually really, really slow.

cowboyd:
Also, should we disable comment integration with the chat? It seems a bit noisy.

taras:
Yeah, I don't know how you did it.

cowboyd:
It's under the Settings > Webhooks

I removed issues comments.... now we know :)

PR #339, opened 2019-03-22
Iterator objects should not be wrapped in arrays

taras:
Array Microstates are iterators. It would be helpful to allow value to be iterators as well. 

Currently, when an array microstate is created with an iterator value, Microstates wraps that value in an array. This messes up the value and prevents the iterator from being used. This PR changes the initializer to detect if the value defines a `Symbol.iterator` property and interprets these as an array.

### DIFF
--- a/src/types/array.js
+++ b/src/types/array.js
@@ -18,7 +18,7 @@ export default parameterized(T => class ArrayType {
   initialize(value) {
     if (value == null) {
       return [];
-    } else if (Array.isArray(value)) {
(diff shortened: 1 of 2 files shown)


taras:
We need additional test coverage around iterators as a value. This is working because spread operator implicitly relies on the iterator. If the implementation of transition was changed, then we might not notice a regression.

taras:
If someone has an expectation that their value would maintain it's type after a transition, then our current implementation would give them grief. With that said, there are very few people who actually think about this so I can't really imagine someone encountering this problem for the foreseeable future.

PR #341, opened 2019-03-26
WIP: Add CircleCI job to publish tagged version to npm

taras:


taras:
> Seems like this would be an ideal candidate for githubs actions

I'm not sure if it's possible to do this with GitHub Actions without creating a custom action.

cowboyd:
This might be true, but there are already actions that you should be able to delegate most of the heavy lifting to https://github.com/actions/npm

cowboyd:
As of v0.15.1 releases are fully automated 🎉

PR #350, opened 2019-04-14
Use new `entries` prop in Object microstate docs

TeknoFiend:
Reflects breaking change in 0.14.0  https://github.com/microstates/microstates.js/pull/338


cowboyd:
We should probably indicate that most often in your code you’ll be using iterations on objects

PR #352, opened 2019-04-16
Added find to query

minkimcello:
This PR introduces the find function inside query.

Currently, we can invoke a `filter` function on a microstate and grab the first item of the resulting array in efforts to mimic the functionality of `find`:
```js
import { create, filter } from 'microstates';

let arr = create(Array, [1, 2, 3, 4, 3])
let filtered = filter(arr, num => num == 3) // [3, 3]
let result = filtered[0] // 3
```

With `find`, returning the desired result becomes simpler:
```js
import { create, find } from 'microstates';

let arr = create(Array, [1, 2, 3, 4, 3])
let find = find(arr, num => num == 3) // 3
```

The `find` function is written in such that at the first instance of the function resulting true will stop the loop and return the result; whereas, `filter`, `reduce`, and `map` would continue its loop until the end.

### DIFF
--- a/src/query.js
+++ b/src/query.js
@@ -14,6 +14,10 @@ export function reduce(iterable, fn, initial) {
   return query(iterable).reduce(fn, initial);
 }
 
+export function find(iterable, fn){
+  return query(iterable).find(fn)
+}
+
 class Query {
   constructor(iterable) {
     if (typeof iterable[Symbol.iterator] === 'function') {
@@ -73,6 +77,17 @@ class Query {
     let result = initial;
     for (let next = iterator.next(); !next.done; next = iterator.next()) {
       result = fn(result, next.value);
+    }
+    return result;
+  }
+
+  find(fn) {
+    let iterator = this.generator();
+    let result;
+    for (let next = iterator.next(); !next.done; next = iterator.next()) {
+      if (fn(next.value)) {
+        return result = next.value
+      }
     }
     return result;
   }
--- a/src/query.js
+++ b/src/query.js
@@ -82,10 +82,9 @@ class Query {
   }
 
   find(fn) {
-    let iterator = this.generator();
-    for (let next = iterator.next(); !next.done; next = iterator.next()) {
-      if (fn(next.value)) {
-        return next.value
+    for (let item of this){
+      if (fn(item)){
+        return item;
       }
     }
   }
--- a/src/query.js
+++ b/src/query.js
@@ -83,13 +83,11 @@ class Query {
 
   find(fn) {
     let iterator = this.generator();
-    let result;
     for (let next = iterator.next(); !next.done; next = iterator.next()) {
       if (fn(next.value)) {
-        return result = next.value
+        return next.value
       }
     }
-    return result;
   }
 
   [Symbol.iterator]() {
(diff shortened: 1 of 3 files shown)


src/query.js
```diff
@@ -77,6 +81,17 @@ class Query {
     return result;
   }
 
+  find(fn) {
+    let iterator = this.generator();
+    let result;
+    for (let next = iterator.next(); !next.done; next = iterator.next()) {
+      if (fn(next.value)) {
+        return result = next.value
```
taras: You don't need to assign result here. You can just return next.value

minkimcello: made the changes. thank you.

src/query.js
```diff
@@ -77,6 +81,15 @@ class Query {
     return result;
   }
 
+  find(fn) {
+    let iterator = this.generator();
+    for (let next = iterator.next(); !next.done; next = iterator.next()) {
```
cowboyd: I think we can even make it simpler if we take into account that the query itself is iterable:

```js
for (let item of this) {
  if (fn(item) {
    return item;
  }
}
```

tests/query.test.js
```diff
@@ -2,23 +2,22 @@
...
 
 describe('A Microstate with queries', function() {
   let todomvc;
-  beforeEach(function() {
-    let start = create(TodoMVC)
-      .todos.push({title: "Take out The Milk"})
-      .todos.push({title: "Convince People Microstates is awesome"})
-      .todos.push({title: "Take out the Trash"})
-      .todos.push({title: "profit $$"});
+  let start = create(TodoMVC)
```
taras: Only define the start variable in the describe block, the assignment should remain in beforeEach.

src/query.js
taras: Can the same be done with reduce? (separate PR)

tests/query.test.js
minkimcello: oh okay

src/query.js
cowboyd: Nice catch! It could totally.

tests/query.test.js
```diff
@@ -49,6 +48,23 @@ describe('A Microstate with queries', function() {
       expect(next.completed).toHaveLength(3);
     });
   });
+
+  describe('testing find function with microstates', function () {
```
cowboyd: We should be able to leave out the fact that this is a test since that is a given. My rule of thumb is that the describe + description should read comfortably a if they were a single sentence: E.g.

```js
describe("finding an element within an iterable Microstate"
  it("can locate the desired item"
  it("returns only the first result"
```

tests/query.test.js
```diff
@@ -65,4 +81,8 @@ describe('Query Array', () => {
   it('can filter a regular array', () => {
     expect([...filter(array, Boolean)]).toEqual([true, true]);
   });
+
+  it('can find a regular array', ()=>{
```
cowboyd: I'd clarify here since the array isn't what's found, it's what's _in_ the array.

```js
it('can find specific items inside a normal array')
```

tests/query.test.js
minkimcello: ohh I see. thank you. that makes much more sense. will do that from now on.

Issue #357, opened 2019-04-24
Control child Microstate creation


cowboyd:
There are many case where we need to specify how a child microstate is created from the parent. A simple case is where you want a microstate to have a default value if none is specified. For example, if I have a counter that I want to start at 1, I would use `create` + the default value.

```js
class Counter {
  count = create(Number, 1);
}
```

Another example is one where we want to pass a value down from one microstate to another. For example, if we need to pass the bluetooth service id down to one of its characteristics, then we'd have to use an initializer.

```js
class Service {
  characteristics = [Characteristic]

  initialize(value) {
    if (this.characteristics.length > 0 && value.characteristics[0].serviceUUID == null) {
      return value.characteristics.map(c => Object.assign({}, c, {serviceUUID: value.uuid}))
    }
    return this;
  }
}
```

This is ugly, and while there are other ways to achieve this, it's not very clear what's going on. Particularly because we are having to fiddle with the parent _value_ in order to achieve a result in the child _microstate_.

So we have two one-off use-cases that leverage existing features that were not really created explicitly for those reasons, but we have to use them because they're kinda just laying around.

What if instead we had a _general_ mechanism for determining how microstates are related to each other, and what is better for a general interface than a function? What if we had a function called `relationship` that was used to specify how to traverse a property from a parent microstate to a child microstate? We could use this for both of the use cases above.

To set a default value:

```js
import { relationship } from 'microstates';

class Counter {
  count = relationship((counter, value) => create(Number, value || 1));
}
```

the `relationship` function receives the parent instance (`counter instanceof Counter`) and the value of the relationship, and is expected to return a microstate representing the relationship.

We can use this low-level function to write a "macro" relationship like `defaults` that does what we did above, except more intent-fully.

```js
import { relationship } from 'microstates';

function defaults(Type, defaultValue) {
  return relationship((parent, value) => create(Type, value == null ? defaultValue : value));
}

class Counter {
  count = defaults(Number, 1);
}

```

We can also use this low-level mechanism for copying values down the tree:

```js
class Service {
  characteristics = relationship((service, values) => create([Characteristic], values.map(v => append(v, { serviceUUID: service.uuid });
}
```

Again though, we can use a "macro" relationship to define this cleanly:

```js

function copyIntoEach(Type, mappings) {
  return relationship((parent, values) => {
    let copies = Object.keys(mappings).reduce((copies, key) => append(copies, {[key]: valueOf(parent)[key])
[...]

PR #358, opened 2019-04-25
Abstract relationships.

cowboyd:
See #357 for full details.

How to contextualize a child microstate has been a constant challenge. In other words, sometimes you want a child to take on a constant value. Sometimes you want it to take on a default value, and sometime you might want it to be completly polymorphic on what value it takes on depending on the parent.

Unfortunately, this is not possible with our current way of expressing relationships with is either with a DSL:

```js
class Counter {
  count = Number;
}
```

or with an explicitly created microstate:

```js
class Counter {
  count = create(Number, 1);
}
```

The problem is that there is no way to expand the concept of what gets created if you want to do something fancier, like copy values down the graph without resorting to intializer hacks.

This introduces a very thin concept of a _relationship_ between a parent microstate and a child microstate. The responsibility of the traversal is to take the value at the location of the target microstate and return a specifier for the child microstate.

To do this, we introduce the concept of a "Cell" which is a Type, value pair `{ Type, value }`. This is the minimal about of information required to specify a microstate. This is super light-weight because it does not require the overhead of actually `create`-ing a microstate.

It is the responsibility of a relationship to resolve a value into a cell.

> Note: in future microstates, we will use `{ Type, path }` as a cell
  specifier.

The relationship class is organized monadically so that it can be mapped, flatMapped, etc... and any relationship can be extended using `map` and `flatMap`

For example, to implement a hypothetical `child` helper which creates a specifies a substate of a given type and default value, it could look like this.

```js
function child(Type, defaultValue) {
  return relationship()
    .map(({ value }) => value != null ? {Type, value} : {Type, value: defaultValue })
}
```

This pull request does not introduce any actual changes built on the `relationship` function, but just exposes it so that we can experiment with it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- Fine grained control of how related Microstates are instantiated
+  via abstract relationships
+  https://github.com/microstates/microstates.js/pull/358
+
 ## [0.14.0] - 2019-03-27
 
 ### Breaking
--- a/index.js
+++ b/index.js
@@ -3,4 +3,5 @@ export { default as from } from './src/literal';
 export { map, filter, reduce, find } from './src/query';
 export { default as Store } from './src/identity';
 export { metaOf, valueOf } from './src/meta';
+export { relationship } from './src/relationship.js';
 export * from './src/types';
--- a/src/microstates.js
+++ b/src/microstates.js
@@ -1,8 +1,9 @@
 import { append, stable, map } from 'funcadelic';
-import { set } from './lens';
+import { set, view, At } from './lens';
 import { Meta, mount, metaOf, valueOf, sourceOf } from './meta';
 import { methodsOf } from './reflection';
 import dsl from './dsl';
+import { Relationship, relationship } from './relationship';
 import Any from './types/any';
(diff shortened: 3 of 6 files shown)


cowboyd:
One thing I was thinking was that instead of using `(value, parent) => Cell` as the arguments, we could use an object so that we can make it rich with metadata.

```js
class Edge {
  get targetValue() {
    return view(At(this.name), this.parentValue);
  }

  get parentValue() {
    return valueOf(this.parent);
  }
   constructor(parent, key) {
     this.parent = parent;
     this.name = name;
   }
}
```

And then the relationship traversal would receive an `Edge` object:

```ts
type Traversal = (edge: Edge) => Cell
```

cowboyd:
Here is a branch using an `edge` argument to the traversal function, rather than passing `(value, parent)` separately. This lets you have ready access to handy things like the location's `value` as well as its `parentValue` https://github.com/microstates/microstates.js/blob/cl/abstract-relationships-using-edges/src/relationship.js#L40-L58

taras:
Edge looks like the way to go because it gives us a richer object to use in the future.

cowboyd:
@taras This PR now uses the `Edge` object as the object to resolve relationships, and also exports the `relationship()` function, and should be good to go w.r.t beta.

taras:
Here is a TypeScript definition for relationships https://gist.github.com/cowboyd/5fd38eb194fc362457909d9dd762f15e

PR #360, opened 2019-04-26
Add tests for module exports.

cowboyd:
The index is uncovered, and it's not clear what our public API is.

This adds some smoke tests to ensure that we don't accidentally remove functions if we re-arrange our exports.

### DIFF
--- a/tests/query.test.js
+++ b/tests/query.test.js
@@ -2,7 +2,7 @@
 import expect from 'expect';
 import { create } from '../src/microstates';
 import { valueOf } from '../src/meta';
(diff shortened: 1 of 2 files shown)


tests/query.test.js
```diff
@@ -2,7 +2,7 @@
 import expect from 'expect';
 import { create } from '../src/microstates';
 import { valueOf } from '../src/meta';
-import { map, filter, reduce, find } from '..';
+import { map, filter, reduce, find } from '../src/query';
```
cowboyd: This was causing breaking when running the tests locally on node 10.

PR #364, opened 2019-05-02
Consuming lens implementation from @microstates/lens

taras:
In the process of building Microstates, @cowboyd wrote our own implementation of lenses. It turned out to be super tiny. I published it as a separate package available as [@microstates/lens](https://github.com/microstates/lens). This PR removes the local implementation in favour of the external package.


cowboyd:
I'd like to make some changes to the lens library first that I've been meaning to make for a really long time. For example, the `childAt` function isn't really related to the `defineChildren` function, and so while the former is critical to lenses, the later isn't. Been meaning to break them up for a while.

Issue #264, opened 2018-11-08
Array microstate functions vs. Query functions


epikhighs:
I was wondering what the difference is b/t using an Array microstate's map, filter and reduce functions VS. the map, filter, reduce found in the Query class?  I found this snippet of info (https://github.com/microstates/microstates.js#mapmicrostate-fn-microstate), but I don't understand why I couldn't just Array.map and get the same result?

From the TodoMVC project, the Array.filter and Query.filter seem to be used interchangeably.  

https://github.com/microstates/todomvc/blob/master/todomvc.js#L35

https://github.com/microstates/todomvc/blob/master/todomvc.js#L96

taras:
That's a great question. 

**Methods on the Microstate are transitions, when invoked they'll create a new microstate with the result of that operation.** 

```js
import { map, create } from 'microstates';

let arr = create(Array, [1, 2, 3]);

arr.map(i => i.increment());
// Microstate<Array>[2, 3, 4]
```

In this example we will create a new microstate with `i => i.increment()` operation applied to each element in the array. It'll return a new microstate with the result of this operation.

**Query functions operate on collections of microstates but don't transition the microstate itself.**

```jsx
import { map, create } from 'microstates';

let arr = create(Array, [1, 2, 3]);

<ul>
 {map(arr, i => <li key={i.state}>{i.state}</li>)}
</ul>
```

In this example, we're mapping over array of microstates and creating components. We don't want to cause transitions while trying to create components, so we use the query. 

This difference becomes even more important when used in a Store. Transitions in a store will cause a re-render cycle, so you have make sure that you use queries.

Let me know if you have more questions. Always happy to answer them.

cowboyd:
This is a very natural source of initial confusion, and is probably worth addressing explicitly in the README.

epikhighs:
Hi guys.  Thanks for the clarification. 

> Methods on the Microstate are transitions

I'm guessing this means the "methods on the **Array** microstate are transitions"?  

Generally speaking, methods in a given microstate aren't required to be transitions right? So in the todoMVC example referenced in the OP,  both `completed()` and `clearCompleted()` are methods on a microstate, but clearly `completed()` is not a transition. Or do ES5 `getters` have some special treatment/usage in microstates?

```
get completed() {
    return filter(this.todos, todo => todo.state.completed);
  }
clearCompleted() {
  return this.todos.filter(({ completed }) => !completed.state);
}
```

[...]

Issue #261, opened 2018-10-31
Updating embedded microstates inside Store


SkaterDad:
Good afternoon!

I first heard of Microstates on your podcast a while back, and the ideas really appeal to me.  I've finally found time to experiment with the library, and have run into some unexpected behavior.

Thanks in advance for any help or instruction you can provide!

**Setup**

I created an `AppState` model which embeds the `Person` and `Car` models from your README, and created a `Store` from it.

I then hooked up that store to re-render a small VDOM library called `superfine`.  The view contains some buttons which reach into the model and update some nested microstates.  I believe I hooked it up correctly, based on what I saw in the `@microstates/react` library.

```js
class AppState {
  person = Person
  car = Car
}
// render is a function that takes a state and invokes a v-dom diff/patch
Store(create(AppState, initialState), render)
```

**What happens:**

Invoking transitions on the "car" microstate will sometimes reset the "person" microstate, and vice versa when viewed from the parameter passed into the Store's `observe` callback. 

**How to reproduce:**

1.  Click the button to update "person" name field, and it works fine.
2.  Click the button to update the "car" name, and it updates, but resets the "person" to the initial state.
3.  Click the "set designer age to 1" button, and car name resets to initial state.
4.  Eventually, click lots of times in random order to get all 3 state changes to take effect at once.

I also tried this with the React `<State>` component with similar results.

Example sandbox:  https://codesandbox.io/s/pj7ry972xj
Simpler React version: https://codesandbox.io/s/w0qr0nypk8

_Full code snippet:_
<details>

```js
import { h, patch } from "superfine"
import { create, Store, valueOf } from "microstates"

// MICROSTATES
class Person {
  name = String
  age = Number

  get state() {
    return valueOf(this)
  }

  get view() {
    return personView(this)
  }
}

class Car {
  designer = Person
  name = String

  get state() {
    return valueOf(this)
  }

  get view() {
    return carView(this)
  }
}

class AppState {
  person = Person
  car = Car

  get state() {
    return {
      person: this.person.state,
      car: this.car.state,
    }
  }
}

// VIEW
function personView(person) {
  return h(
    "p",
    {
      oncreate: () => console.log("Created person view: " + person.name.state),
      onupdate: () => console.log("Updated person view: " + person.name.state),
    },
    `${person.name.state} is ${person.age.state} years old.`
  )
}

function carView(car) {
  return h(
    "div",
    {
      oncreate: () => console.log("Created car view: " + car.name.state),
[...]

PR #371, opened 2019-09-24
Setup Transparent NPM Publishing

minkimcello:
# Purpose
This PR is to setup **Transparent NPM Publishing** for `microstates/microstates.js`.

# Approach
Screen Shot 2019-10-02 at 1 48 55 PM

## Publish Preview Workflow
When a pull request is created, it will first confirm that the pull request has access to its secrets for `NPM_AUTH_TOKEN`. Then the changes of the pull request will publish to NPM and if successful a comment will be generated on the pull request with instructions on how to access the NPM package.

## Publish Release Workflow
This workflow triggers on each push/commit to the repository and is meant to run for automated commits when a pull request is merged. (The workflow will halt at the beginning and return an error if it cannot detect a pull request from the commit message.) If it can confirm that it's being run on the commit of a merge, it will fetch the labels of that pull request, apply the necessary labels for updating package version, commit/push package.json back to the repository, and then publish the package.

# Setup Checklist
Below are the steps necesseary to start utilizing these two workflows:
- [ ] approve [PR #8](https://github.com/thefrontside/actions/pull/8) for `thefrontside/actions`
- [ ] add `NPM_AUTH_TOKEN` from npmjs in secrets of this repository
- [ ] uncomment branches in both workflows

# Learning
This setup has been adapted and tested in [Georgia](https://github.com/taras/georgia/tree/release-1.0.0).

The actions are coming from [thefrontside/actions](https://github.com/thefrontside/actions).


.github/workflows/pr-create.yml
```diff
@@ -0,0 +1,41 @@
+name: TNP PR Created
```
cowboyd: Let's rename this "Publish Preview" since this is what the workflow does. "pr created" is the event that triggers the workflow, but it is not the workflow itself.

.github/workflows/pr-create.yml
```diff
@@ -0,0 +1,41 @@
...
+
+on:
+  pull_request:
+    branches:
+      - release-*
```
cowboyd: Don't we want `master` to be in this list as well?

.github/workflows/pr-merge.yml
```diff
@@ -0,0 +1,69 @@
+name: TNP PR Merged
```
cowboyd: In the someway we called the other workflow "publish preview", let's call this 'publish release"

.github/workflows/pr-create.yml
minkimcello: Updated

.github/workflows/pr-merge.yml
minkimcello: Updated

.github/workflows/pr-create.yml
minkimcello: Yeah, that'd be a good idea. I updated it but would it be better if we included **all** pull requests from every branch?

jbolda:
Are you planning on versioning the `thefrontside/actions/*`? I noticed that you are currently using `master`.

minkimcello:
> Are you planning on versioning the `thefrontside/actions/*`? I noticed that you are currently using `master`.

@jbolda I was putting that off but I should probably get around to it. :P

.github/workflows/pr-create.yml
cowboyd: Hmmm.... that's actually a great point. As long as we're judicious about what branches we merge to, then maybe this restriction is pointless?

minkimcello:
> Another thing occurred to me as a point of due-diligence. Because we're allowing publishing from this repository's actions. How does this work with forks of this repository? I wouldn't want someone to be able to fork this repo, submit a PR for one of their own branches, and be allowed to publish something to NPM

@cowboyd They won't have access to *our* `NPM_AUTH_TOKEN` unless they're merging onto our repository. Would it be an issue if they used the workflow to publish to their own NPM account?

> Also can we name the files publish-preview.yml and publish-release.yml?

Done.

minkimcello:
@cowboyd [This](https://github.com/r-lib/ghactions/issues/262) is what I've found on the subject. It's from June so not sure how it might've changed since then but here:
> pull requests from a fork won't show the check results in the PR view (see #261 (non-beta), #263 (beta)).
Presumably, this is because an action on a PR gets run against the fork, not base, for security reasons (bad actors could make main.workflow reveal secrets).
If the fork creator has the beta, s/he can at least see the actions run there. (Actually, s/he can't because first need to delete/recommit main.workflow as per #264).

jbolda:
I think they changed it a bit since then:
https://help.github.com/en/articles/events-that-trigger-workflows#pull-request-events-for-forked-repositories
>When you open a pull request from a forked repository, the pull request opens on the base repository by default. In this scenario, GitHub sends the pull_request event to the base repository and no pull request events occur on the forked repository because the webhook configuration of the original base repository doesn't apply to forked repositories.
You can create a new workflow or configure webhooks in the forked repository. For more information, see "Configuring a workflow" or "Creating webhooks" in the GitHub Developer documentation.
The permissions for the GITHUB_TOKEN in forked repositories is read-only. For more information about the GITHUB_TOKEN, see "Virtual environments for GitHub Actions."
>

minkimcello:
@jbolda I just forked `taras/georgia` and tried creating a PR against `release-1.0.0` of base and the publishing step returned a 401 unauthorized error. It's not conclusive but I would guess that if the NPM_TOKEN was that easily accessed, it'd be a huge security flaw?

jbolda:
Oh, I forgot to add to my previous, that doesn't apply to secrets set in the repository. So I think your NPM_TOKEN is safe from others, but it seems like this workflow wouldn't publish anything if someone submits a PR from a forked repository (which is maybe intentional?). I just happened to deep dive this stuff this weekend; [setup CI on one repo](https://github.com/jbolda/finatr/pull/151) and was looking into how secrets function for [another repo](https://github.com/jbolda/gatsby-source-airtable).

https://help.github.com/en/articles/virtual-environments-for-github-actions#creating-and-using-secrets-encrypted-variables
>With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

minkimcello:
@cowboyd Okay so I spent the morning experimenting with a forked repository of `taras/georgia` and here's what I found:

## Findings
- Forked repositories can run their own workflows within their repository but will have to update package name (so that it doesn't try and publish to someone else's package) and provide their own NPM token.

- When a fork creates a PR against the repository it forked from, it runs the workflow configured in their own commit (which can be tinkered with) and the PR does not have access to the secrets they defined in the forked repository. Meaning, as @jbolda pointed out, none of the secrets are available except for the GITHUB_TOKEN which is read-only. In the context of our current workflows, the workflow can't generate a comment with Danger and obviously can't push to the repository.

- I was thinking of perhaps creating a new workflow with a filter based on the actor for forked repositories so that it can automatically update package name with actor and publish using their own token but that won't work because it doesn't have access to the secrets defined in their own fork. I also thought of having a workflow that triggers after someone with admin privileges of the main repository labels it with an "okay" but we still run into the same issue.

## Conclusions & Possible Solutions
- Publishing can't be done from a fork. Can be done *on* a fork but they will need to configure for their own setup.

- We could slap on an actor filter on the two current workflows. Although it can be modified from a fork it'll return an error down the road so it's just a prettier way of preventing the workflow from running.

- So if we conclude that preview-publishing must be forfeited on forked pull requests, we now deal with the issue of release publishing. Currently publish-release workflow is run *on* the PR because of my oversight. In the case when a fork PR is merged, we don't have a way of publishing that yet (because right now it'll just return an error).

  - We could reincorporate what we had before which is to trigger the publish-release workflow on commits of the base which means reverting back to filters that search for \#(0-9)*\. We could attach a filter so that this method only runs when a pr is merged from a fork which would distinguish the various type of workflows but migrating all publish-release workflows in the old way would make maintaining the use of these workflows easier. 

    - I rechecked the payload again and commits from a merge does not contain any information on the pull request except from the commit message.

## Conclusion Conclusion
Here are two scenarios I can think of at the moment:
- Scenario 1:
  - `publish-preview` with filter to prevent forks
  - `publish-release` with filter to prevent forks
  - `publish-release-for-forks` triggered on commits to main repository with filter specifically for forks

- Scenario 2:
  - `publish-preview` with filter to prevent forks
  - `publish-release` run on `push` of base repository after merge commits = filter on commit messages.

In both scenarios, if forked repositories want to test out their preview package, they need to run it on their own repository.

And in both scenarios we need to readjust the label fetching action to retrieve it from the appropriate source; scenario 1 would mean we need to create a second action, and scenario 2 would mean we modify the one we have now for our use case.

jbolda:
So I have been mulling over how best to deal with forked repos in the grand scheme of things, and I feel it might be worth expanding on @minkimcello conclusions. I have done this in the past, and it feels like the best way to approach it considering the obstacles.

To frame things a bit here, the purpose of this system is to make publishing smooth. When we publish a major or minor, more effort may intentionally be required to be put forth. With patches though, they should be relatively trivial to publish after test cases and such pass. The most interesting bit of this is publishing "dev" versions of in-progress or partially completed code changes. I would expect that this feature will mostly be used by collaborators of the project (in which case, this will work just fine), and that those coming into the repo and submitting a PR off of a fork are not expecting to get version for their commits until it is merged.

Expanding on that assumption, I would expect there is limited value in the creation of many dev versions as a PR progresses. I think the real value here is in being able to minimize friction of testing long-running or involved code changes. This could arguably be best done on a specific feature branch. If we adjust our scope to focus around feature branches instead of PRs, I think we can strike a good balance of ease, control and working around fork repository issues. It would require that some procedures be documented for creation of branches to trigger this workflow.

A few simple examples may help.

With the case of the core contributor, they may create a PR but not need or want the noise of dev versions being created. They can create a branch name without [yet to be determined feature branch naming scheme], and proceed as normal.

If there is the expectation that this will be long-running, a branch can be created where a publish new dev version action is run on each commit. Then as work progresses, either through PRs that get merged or direct commits to the branch, we can publish a dev version which includes the name of the branch and an incrementing number (which makes it easier for someone on the outside to understand how to stay abreast of this sequence of versions).

In the case of a fork PR coming in that turns out to be long-running, a new branch can be created, the PR can be switched to target that new feature branch, and when it is merged a new dev version will be created. I think this has the side effect benefit of encouraging smaller PRs, and encouraging more in-depth collaboration on these high effort features (where you would especially find value in minimizing the effort for someone to test it out in their own repo).

</end examples>

I am making some assumptions on the ultimate end goal here, but if I am being selfish, I think this might best meet my goals if/when I bring these actions into personal repos.

cowboyd:
@minkimcello I'm ok with either scenario, but it sounds like (2) is less complex. As for publishing a preview from a fork. I guess we could do something where if they have an NPM token, and they want to do it (not sure how they'd specify) but it could publish within their user namespace on NPM. So if I am `cowboyd` on npm (which I am), then from a fork, I could publish `@cowboyd/microstates-ab7de67` provided I had my `NPM_TOKEN` entered in, and maybe some other opt-in mechanism?

I think for the time being, it's perfectly acceptable to have no previews on forks as long as we have them for every PR in the repo, and that releases happen on PR merges from forks.

@jbolda I think we can be more conservative about which versions get published after we have the baseline in place. Our goal here is to be as transparent as possible about what is happening with our OSS repos so that by default our users can take our latest work and integrate it immediately. I could see having a label like "no-preview" or filtering if the branch name ends in `-no-preview`, but out-of-the-box, every PR that happens in the repo will have a preview associated with it.

cowboyd:
As a matter of fact, the philosophical underpinnings of our approach are the subject of our most recent podcast https://frontside.io/podcast/transparent-development

minkimcello:
@cowboyd 
> I'm ok with either scenario, but it sounds like (2) is less complex.

I'll make the changes for scenario 2.

> As for publishing a preview from a fork. I guess we could do something where if they have an NPM token, and they want to do it (not sure how they'd specify) but it could publish within their user namespace on NPM.

I need to double check what's available in the payload of a fork but forking can trigger an actions workflow so I'll see what kind of solution I can come up with. I can also update the readme and create an error message for when publishing is attempted without an npm token.

> So if I am `cowboyd` on npm (which I am), then from a fork, I could publish `@cowboyd/microstates-ab7de67` provided I had my `NPM_TOKEN` entered in, and maybe some other opt-in mechanism?

This is what I had in mind too. I should be able to modify the package name with the actor right before it publishes so that the change to the name is discarded if and when it's merged to base.

cowboyd:
Sounds good, but I don’t think forked Previews should be a blocker for us.

jbolda:
@cowboyd got it. Yeah, this setup should work wonderful for members of the repo. My thoughts RE: branches were more of an additional workflow for branches to deal with forked repositories. For my specific case, I _also_ have to deal with secrets being required to run tests as well so forked repositories become a real pain to adequately approve. Agreed that it doesn't make sense to be a blocker though.

.github/workflows/publish-preview.yml
```diff
@@ -0,0 +1,46 @@
...
+        NPM_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
+
+  comment: 
+    name: Post Instruction Comment
+    runs-on: ubuntu-latest
+    needs: publish
+    steps:
+    - uses: actions/checkout@v1
+    - name: Post Comment
+      uses: thefrontside/actions/post-instruction-comment@master
```
cowboyd: This action should be named "post npm usage instructions comment" or the like because as-is it feels like you could pass any block to it, but it is npm specific.

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,75 @@
...
+jobs:
+  filter:
+    name: Filter for Merges
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    # stage for adding to thefrontside/actions
+    - uses: taras/georgia/.github/actions/filter-by-commit@release-x
+
+  tests:
```
cowboyd: Now I'm wondering if we should be blocking the publish on tests because this PR should not have been merged to the branch without passing the tests. Until we're running our CI in GitHub, I'm thinking it's best not to actually run the tests since there will be invariable divergence between the Circle environment and the GitHub environment.

.github/workflows/publish-preview.yml
minkimcello: updated

.github/workflows/publish-release.yml
minkimcello: we can take the tests out and put it back in when it's more appropriate.

oh and you can track my progress [here](https://github.com/minkimcello/microstates.js) @ release-0.

minkimcello: and actions [here](https://github.com/minkimcello/actions)

minkimcello:
Updated PR. Should I squash the commits now or just have it squashed when we merge?

PR #388, opened 2019-12-04
Update action source from master to v1.1

minkimcello:
We're going to be introducing some breaking changes and so we created a "save point" by tagging and releasing the current state of our `thefrontside/actions` as `v1.1`. There might be other bugs but going from `master` to `v1.1` in this repository's workflows should not be any different as `master` and `v1.1` are identical at the moment.

### DIFF
--- a/.github/workflows/publish-preview.yml
+++ b/.github/workflows/publish-preview.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: NPM Publish Preview
-      uses: thefrontside/actions/publish-pr-preview@master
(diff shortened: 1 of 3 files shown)


frontsidejack:
A preview package of this pull request has been published with the tag `mk_action-version`.
You can try it out by running the following command:
```bash
$ npm install microstates@mk_action-version
```
or by updating your package.json to:
```bash
{
  "microstates": "mk_action-version"
}
```
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against e2098136a39ab8c666e81d29a35a2064cdaf5c7a
</p>

